PR #19, opened 2021-12-30
FontIcon property added to NavigationItem

enginkirmaci:


Pomianowski:
Hey, thanks for the PR. I guess if we give the option to add an icon from the font, there should also be an option to select the font itself.

enginkirmaci:
> Hey, thanks for the PR. I guess if we give the option to add an icon from the font, there should also be an option to select the font itself.

It's a good idea, I added fontfamily too.

PR #62, opened 2022-02-15
Supports net45 instead of net46 and net47

walterlv:
I've checked all the source code and found that we didn't use any features more than net45, so I think we can support net45 without any further burden.

BTW, net451/net452/net46/net461/net462/net47/net471/net472/net48 projects support the net45 NuGet package so that we don't need to compile a library targeting above net45. However, because net47/net471/net472/net48 support ValueTuple natively so it deserves to stay still in the project to reduce dependency.


MajorXAML:
Why?

https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

walterlv:
Windows 7 without SP1 supports net452 as its latest .NET Framework, but there are still some users using this version of system. This application below need to run on Windows 7 (even without SP1):

* https://github.com/kkwpsv/WindowDebugger/pull/4

MajorXAML:
Well, some people are still using windows xp... these users should take care of the security of their systems and upgrade their os :)

walterlv:
OK, I'll compile the net45 version in private for this app above. Thanks!

MajorXAML:
It is worth waiting for a response from the author of the library, I just gave my opinion :)

pomianowski:
Windows 7 has finished its life and will not be supported.
NET Framework is only added for backwards compatibility, the target for the library is .NET 6.

PR #66, opened 2022-02-18
Fix the client area margin at High DPI devices.

walterlv:
## What does this pull request do?

In different DPIs, the margins of the client area are different. What is strange is that the values of the different margins are not a linear scale value of the DPI factors. If we use a constant value as the client area margin, the WPF will auto-scale it by using a linear function with the DPI value which is NOT CORRECT. What I'm doing is to fix the incorrect auto-scaling behavior of the client area margin.

The actual margin value parts in WPF units in different system DPI values:

| DPI  | WindowNonClientFrameThickness           | PaddedBorderThickness |
| ---- | --------------------------------------- | --------------------- |
| 100% | 4,4,4,4                                 | 4,4,4,4               |
| 125% | 3.2,3.2,3.2,3.2                         | 4,4,4,4               |
| 150% | 3.3333,3.3333,3.3333,3.3333             | 4,4,4,4               |
| 175% | 2.8571428,2.8571428,2.8571428,2.8571428 | 4,4,4,4               |
| 200% | 2.5,2.5,2.5,2.5                         | 4,4,4,4               |

If we do not match the correct client area margins, the user will see a space between the minimized/maximized/close button and the screen edges when the window is maximized.

## How does the source code work?

Thanks to my friend @DinoChan which is also a Microsoft MVP in Windows Development Area. He researched this behavior of the client area of Windows and teach me this by publishing his blog post. I've used this knowledge by writing a custom control named `ClientAreaBorder` which is the source code you can see in this pull request. This code has been tested in a widely published popular software for more than a year and works fine. The original source code is here:

https://github.com/walterlv/Walterlv.Packages/blob/master/src/Themes/Walterlv.Themes.FluentDesign/Controls/ClientAreaBorder.cs

## Extras

A screenshot of the fix of the margin of 150% DPI:

image

## Related links

* [[WPF 自定义控件]使用WindowChrome的问题 - dino.c - 博客园](https://www.cnblogs.com/dino623/p/problems_of_WindowChrome.html)

### DIFF
--- a/WPFUI/Controls/ClientAreaBorder.cs
+++ b/WPFUI/Controls/ClientAreaBorder.cs
@@ -1,0 +1,103 @@
+﻿#nullable enable
+using System;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Shell;
+
+using static WPFUI.Win32.User32;
+
+using WPFUI.Common;
+
+namespace WPFUI.Controls
+{
+    /// <summary>
+    /// If you use <see cref="WindowChrome"/> to extend the UI elements to the non-client area, you can include this container in the template of <see cref="Window"/> so that the content inside automatically fills the client area.
+    /// Using this container can let you get rid of various margin adaptations done in Setter/Trigger of the style of <see cref="Window"/> when the window state changes.
+    /// </summary>
+    public class ClientAreaBorder : Border
+    {
+        private const int SM_CXFRAME = 32;
+        private const int SM_CYFRAME = 33;
+        private const int SM_CXPADDEDBORDER = 92;
+
+        private Window? _oldWindow;
+        private static Thickness? _paddedBorderThickness;
+        private static Thickness? _resizeFrameBorderThickness;
+        private static Thickness? _windowChromeNonClientFrameThickness;
+
+        /// <inheritdoc />
+        protected override void OnVisualParentChanged(DependencyObject oldParent)
+        {
+            base.OnVisualParentChanged(oldParent);
+
+            if (_oldWindow is { } oldWindow)
+            {
+                oldWindow.StateChanged -= Window_StateChanged;
+            }
+
+            var newWindow = (Window?)Window.GetWindow(this);
+            if (newWindow is not null)
+            {
+                newWindow.StateChanged -= Window_StateChanged;
+                newWindow.StateChanged += Window_StateChanged;
+            }
+
+            _oldWindow = newWindow;
+        }
+
+        private void Window_StateChanged(object? sender, EventArgs e)
+        {
+            var window = (Window)sender!;
+            Padding = window.WindowState switch
+            {
+                WindowState.Maximized => WindowChromeNonClientFrameThickness,
+                _ => default,
+            };
+        }
+
+        /// <summary>
+        /// Get the system <see cref="SM_CXPADDEDBORDER"/> value in WPF units.
+        /// </summary>
+        public Thickness PaddedBorderThickness
+        {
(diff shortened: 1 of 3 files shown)


MajorXAML:
I noticed that the window "After this pull request" has no border 🧐

walterlv:
@MajorXAML neither has borders. There is a bug where borders appear to be present at 150% DPI.

MajorXAML:
According to WinUI 3 the border of window should be 1 pixel thick and use `SurfaceStrokeColorDefault` color
image
image


So the bug is in the lack of border 🤔

walterlv:
@MajorXAML We are solving the maximized window issue. The 1 pixel only exists in a normal window.

pomianowski:
Thanks @walterlv  for your participation in the development and for sending the PR.

blakepell:
Is this by chance in the `1.2.2-prerelease149` on Nuget?  It changes where the border is based where the monitor is located.  Monitor 1 is a standard resolution monitor scaled to 125%, the orange border in the 2nd screenshot is at the top.  Monitor 2 is a standard resolution monitor scaled to 125% and it has no top border (looks right) but has a 1 pixel right border that's the window highlight color.  Monitor 2 (4K, 150% scaling looks fine).

image

image

walterlv:
@blakepell Can I know the DPI awareness of your app?

- System Awareness
- Per-monitor Awareness
- Per-Monitor Awareness V2

I'm trying to reproduce your issue.

blakepell:
@walterlv Thank you for the questions.  They got me on the right track and it's working as expected now.  I realized after I read your quetions that I had an app.manifest in the project but I didn't have it listed in the csproj so it wasn't using it.  I'll go ahead and share those snippets in case they help anyone else.

```xml
  <PropertyGroup>
    <OutputType>WinExe</OutputType>
    <TargetFramework>net6.0-windows</TargetFramework>
    <Nullable>enable</Nullable>
    <UseWPF>true</UseWPF>
    <ApplicationManifest>app.manifest</ApplicationManifest>
  </PropertyGroup>
```

#### app.manifest

``` xml
<?xml version="1.0" encoding="utf-8"?>
<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
	<assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
	<trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
		<security>
			<requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
				<requestedExecutionLevel level="asInvoker" uiAccess="false" />
			</requestedPrivileges>
		</security>
	</trustInfo>

	<compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
		<application>
			<!-- Windows 10 -->
			<supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
		</application>
	</compatibility>

	<application xmlns="urn:schemas-microsoft-com:asm.v3">
		<windowsSettings>
			<dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
			<dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
			<longPathAware xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware>
		</windowsSettings>
	</application>
</assembly>
```

walterlv:
@blakepell Great!

PR #115, opened 2022-03-26
Added CanMaximize property to Titlebar 

JannesStroehlein:
Hello,

I found it very unelegant to just point the MaximizeActionOverride to an empty action. Like this:
```csharp
private void MaximizeOverwrite(TitleBar tb, Window w) 
{

}
```
So I added a property to the Titlebar class that completely disables maximizing (for me just disabling the maximize button wasn't sufficient, because double clicking still maximized the Window).

I hope this addition is in line with your design principles.

With kind regards

Jannes

PS: I also thought about automatically disabling minimize, maximize, resize and snapping when the WindowStyle is set to ToolWindow.


WPFUI/Controls/TitleBar.cs
```diff
@@ -80,6 +80,13 @@ public class TitleBar : UserControl
             nameof(ShowHelp),
             typeof(bool), typeof(TitleBar), new PropertyMetadata(false));
 
+        /// <summary>
+        /// Property for <see cref="CanMaximize"/>
+        /// </summary>
+        public static readonly DependencyProperty CanMaximizeProperty = DependencyProperty.Register(
+            nameof(CanMaximize),
+            typeof(bool), typeof(TitleBar), new PropertyMetadata(false));
```
pomianowski, line 88: It should be `true` by default

pomianowski (changes requested):

WPFUI/Controls/TitleBar.cs
JannesStroehlein: Done

pomianowski (approved):

PR #119, opened 2022-03-31
Fix raise max button before releases the mouse when use snap.

wherewhere:
I move the "RaiseButtonClick()" to the NCLBUTTONUP message. It will raise button after releases mouse now.
动画

### DIFF
--- a/WPFUI/Common/SnapLayout.cs
+++ b/WPFUI/Common/SnapLayout.cs
@@ -84,12 +84,10 @@ namespace WPFUI.Common
                     break;
 
                 case User32.WM.NCLBUTTONUP:
-                    if (!_isButtonClicked) { break; }
+                    if (!_isButtonClicked) break;
 
                     if (IsOverButton(wParam, lParam))
-                    {
                         RaiseButtonClick();
-                    }
 
                     _isButtonClicked = false;
                     handled = true;
--- a/WPFUI/Common/SnapLayout.cs
+++ b/WPFUI/Common/SnapLayout.cs
@@ -84,15 +84,16 @@ namespace WPFUI.Common
                     break;
 
                 case User32.WM.NCLBUTTONUP:
-                    if (_isButtonClicked)
+                    if (!_isButtonClicked) { break; }
+
+                    if (IsOverButton(wParam, lParam))
                     {
-                        if (IsOverButton(wParam, lParam))
-                        {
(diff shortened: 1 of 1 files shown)


WPFUI/Common/SnapLayout.cs
```diff
@@ -81,6 +83,18 @@ private IntPtr HwndSourceHook(IntPtr hWnd, int uMsg, IntPtr wParam, IntPtr lPara
 
                     break;
 
+                case User32.WM.NCLBUTTONUP:
+                    if (_isButtonClicked)
```
pomianowski, line 87: What about
```c#
if (!_isButtonClicked || !IsOverButton(wParam, lParam)) break;
```

pomianowski (changes requested):

WPFUI/Common/SnapLayout.cs
wherewhere: I think "if" need "{}".

wherewhere: I edit it in GitHub, so I can’t make sure it has no error.

wherewhere:
Any one here?

wherewhere:
No one care about me...

WPFUI/Common/SnapLayout.cs
```diff
@@ -81,6 +83,19 @@ private IntPtr HwndSourceHook(IntPtr hWnd, int uMsg, IntPtr wParam, IntPtr lPara
 
                     break;
 
+                case User32.WM.NCLBUTTONUP:
+                    if (!_isButtonClicked) { break; }
```
pomianowski, line 87:
```c#
if (!_isButtonClicked) break;
```

WPFUI/Common/SnapLayout.cs
```diff
@@ -81,6 +83,19 @@ private IntPtr HwndSourceHook(IntPtr hWnd, int uMsg, IntPtr wParam, IntPtr lPara
...
+
+                    if (IsOverButton(wParam, lParam))
+                    {
```
pomianowski, line 90:
```c#
 if (IsOverButton(wParam, lParam))
    RaiseButtonClick();
```

pomianowski (changes requested):

pomianowski (approved):

Issue #229, opened 2022-06-18
Add Dependency Injection


IvanDmitriev1:
It would me nice if dialog, snackbar and navigation would work with DI.

Two month ago I already integrated dependency injection with this controls,  you can check how it works in general https://github.com/IvanDmitriev1/wpfui

PR #231, opened 2022-06-18
Add dependency injection for snackbar, dialog

IvanDmitriev1:
Closes https://github.com/lepoco/wpfui/issues/229
## Pull request type

- [ ] Update
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Additionally  added [ServicesCollectionExtension](https://github.com/lepoco/wpfui/commit/dee48131b40a4867f46c868b2c5e7a1f69f75b0e) with AddWpfUiServices extension.


pomianowski (changes requested):

src/Wpf.Ui.Demo/App.xaml.cs
```diff
@@ -49,17 +49,14 @@ private async void OnStartup(object sender, StartupEventArgs e)
     /// <summary>
     /// Configures the services for the application.
     /// </summary>
-    private void ConfigureServices(HostBuilderContext context, IServiceCollection services)
+    private static void ConfigureServices(HostBuilderContext context, IServiceCollection services)
     {
+        //Add all wpf services
+        services.AddWpfUiServices();
```
IvanDmitriev1, line 55: Why not just use one method that adds all the related wpf UI services?

IvanDmitriev1: And I suppose I Page Service implementation should be in  wpf ui.

pomianowski: > Why not just use one method that adds all the related wpf UI services?

Because someone may not want to use all of them and because not everyone wants to use Dependency Injection

> And I suppose IPageService implementation should be in wpf ui.

No, because this implementation should be independent of the ServiceCollection provider

pomianowski (approved):

Issue #294, opened 2022-07-21
[Enhancement/New control] Color Picker


CocoCR300:
**Is your feature request related to a problem? Please describe.**
This library and WPF are missing a ColorPicker control.

**Describe the solution you'd like**
I've been working on porting the [WinUI](https://docs.microsoft.com/en-us/windows/winui/api/microsoft.ui.xaml.controls.colorpicker?view=winui-2.8) color picker to WPF for the last few days and I've made a great progress by now, and I would like to contribute to this library with it.

**Describe alternatives you've considered**
...

**Additional context**
...

EddyHaigh:
Is the color-picker completed but missing from the samples gallery?

chucker:
> Is the color-picker completed

No. That PR would have to be reworked before it can get merged.

PR #302, opened 2022-07-24
[Enhancement/New control] Color picker

CocoCR300:
## Pull request type

- [ ] Update
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

As said in #294, we don't have a built in color picker in WPF, nor this library provides one, so here it is!

## What is the new behavior?
A new color picker control, ported from the [Microsoft.UI.Xaml](https://github.com/microsoft/microsoft-ui-xaml/tree/main/dev/ColorPicker) repo.

## Other information
**It is not ready yet**, I didn't found some of the resources used by the controls on the original repo and some work is pending:
- [x] ~~Alpha slider and alpha text input can not be hidden individually (Only by disabling alpha completely).~~
- [x] Color picker color slider and hex input can not be hidden.
- [x] Color picker more button does not show.
- [x] Alpha slider checkered background does not show.
- [x] Color picker slider vertical orientation adds a spacing both at the top and the bottom of the control.
- [x] Large increments on color (done by pressing Ctrl+Any arrow key, while the color spectrum is focused) work only in some cases. The problem is located in the [ColorHelpers.IncrementColorChannel](https://github.com/CocoCR300/wpfui/blob/ec3bd057fe6d182f5d8856e4f00c003140f4c764/src/Wpf.Ui/Common/ColorHelpers.cs#L116) method.
- [x] Fix "third dimension slider" and "alpha slider" tooltips.
- [x] Maybe create new templates for the color picker slider.
- [ ] Maybe improve the way to obtain the name of a color, currently this method does the job: [ColorHelpers.GetNearestNamedColorTuple](https://github.com/CocoCR300/wpfui/blob/ec3bd057fe6d182f5d8856e4f00c003140f4c764/src/Wpf.Ui/Common/ColorHelpers.cs#L47).
- [x] Replace hardcoded strings and color names.
- [x] Maybe other things that I have not tested, and... cleanup, I left some old code commented just in case.

I'll be working on what's left until my vacation ends, after that I will be busy, but hopefully it will be finished by that time

### DIFF
--- a/src/Packages.props
+++ b/src/Packages.props
@@ -2,6 +2,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
     <PackageReference Update="Colourful" Version="3.1.0"/>
+    <PackageReference Condition ="'$(TargetFramework)' != 'net461'" Update="Lepo.i18n" Version="1.2.2"/>
     <PackageReference Update="Microsoft.Toolkit.Mvvm" Version="7.1.2"/>
     <PackageReference Update="Microsoft.Extensions.Hosting" Version="6.0.1"/>
     <PackageReference Update="System.Drawing.Common" Version="6.0.0" />
--- a/src/Packages.props
+++ b/src/Packages.props
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
+    <PackageReference Update="Colourful" Version="3.1.0"/>
     <PackageReference Update="Microsoft.Toolkit.Mvvm" Version="7.1.2"/>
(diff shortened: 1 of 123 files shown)


CocoCR300:
@pomianowski, I was planning on using your Lepo.i18n library for localization, but it does not have support for .NET Framework 4.6.1, which this library does. Is there any reason to not support that .NET Framework version on that library?

CocoCR300:
I am finished, there is still some work to do. I used the Lepo.i18n library even if does not have support for .NET Framework 4.6.1, but in a quick test, I verified that it could. Also, I just added the localized color names for english (US and GB) and spanish (ES and MX), so there is pending work there, but I am going to take a break on these last few days of my free time.
Sorry for no finishing this up and not provide any documentation, but I already took a great amount of my vacation working on another project and this one.

pomianowski (changes requested):

Issue #396, opened 2022-09-20
Can not show MessageBox when Application.Current.MainWindow is null


phamviethuy:
**Describe the bug**  
When Application.Current.MainWindow is null. Can not show a message box and throw an exception 

**Expected behavior**  
Show a message box when Application.Current.MainWindow is null.

**Screenshots**  
![image](https://user-images.githubusercontent.com/42407610/191169227-4b03751d-c2a1-4cbc-828a-4941cf8acffd.png)

phamviethuy:
I created pull request to fix this bug.
https://github.com/lepoco/wpfui/pull/397

PR #397, opened 2022-09-20
fix bug open message box when MainWindow is null

phamviethuy:
fix bug open message box when MainWindow is null

## Pull request type

- [ ] Update
- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes

When MainWindow is null. a message box is not opened.

Issue Number: #396 

## What is the new behavior?

When MainWindow is null. a message box is opened on center screen


src/Wpf.Ui/Controls/MessageBox.cs
```diff
@@ -238,6 +237,18 @@ public void Show(string title, object content)
     //    base.OnContentChanged(oldContent, newContent);
     //}
 
+    private void SetWindowStartupLocation()
+    {
+        if (Application.Current.MainWindow != null)
```
chucker, line 242: Can we do `Application.Current?.MainWindow` instead? The else branch doesn't require `Application.Current` to be set

pomianowski (approved):

PR #521, opened 2023-02-01
Controls update

IvanDmitriev1:
Deleted `Dialog` control.
Changed .editorconfig so that it is the same as in the [dotnet/runtime repository.](https://github.com/dotnet/runtime/blob/main/.editorconfig)
Added folders for `TitleBar`, `MessageBox`, `ContentDialog`, `VirtualizingWrapPanel`, `NumberBox `controls.
Added small fixes.

## Pull request type

- [X] Update
- [ ] Bugfix
- [ ] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build related changes


pomianowski (changes requested):
Hey, thanks for organizing the cleanup.

For the namespace to work properly, **Properties\AssemblyInfo.cs** needs to be updated. Same for **VisualStudioToolsManifest.xml**.

As for `IDialogService`, I find it useful and think that it should be rewritten for `DialogControl`.

IvanDmitriev1:
Ok.

IvanDmitriev1:
> For the namespace to work properly, **Properties\AssemblyInfo.cs** needs to be updated. Same for **VisualStudioToolsManifest.xml**.

Actually I didn't change namespace, they all have the `Wpf.Ui.Controls` namespace.

pomianowski:
> > For the namespace to work properly, **Properties\AssemblyInfo.cs** needs to be updated. Same for **VisualStudioToolsManifest.xml**.
> 
> Actually I didn't change namespace, they all have the `Wpf.Ui.Controls` namespace.

I'm not a fan of the solution when the namespace is not the same as the directory tree.

IvanDmitriev1:
At the same time, all controls should have the namespace `Wpf.Ui.Controls`, isn't that right?

pomianowski:
> At the same time, all controls should have the namespace `Wpf.Ui.Controls`, isn't that right?

No need, namespace is added to **AssemblyInfo**. 

https://github.com/lepoco/wpfui/blob/87f203ce2de3aacb2b15d8de6fd8c326bf7575d9/src/Wpf.Ui/Controls/Navigation/NavigationView.Parent.cs#L11
https://github.com/lepoco/wpfui/blob/87f203ce2de3aacb2b15d8de6fd8c326bf7575d9/src/Wpf.Ui/Properties/AssemblyInfo.cs#L20

IvanDmitriev1:
> No need, namespace is added to **AssemblyInfo**.

But it for XAML not for c#.

pomianowski:
> > No need, namespace is added to **AssemblyInfo**.
> 
> But it for XAML not for c#.

It doesn't matter what the namespace is, it could even be **Wpf.Ui.Xyz** and everything should still work

IvanDmitriev1:
> Unless you mean a naming convention to keep everything neatly in one place. Well, so far I've decided that we should stick to the directory tree, but if everything in **Controls** is more convenient and in line with what is in the main WPF, we can consider it that way.

Yep

pomianowski:
> > Unless you mean a naming convention to keep everything neatly in one place. Well, so far I've decided that we should stick to the directory tree, but if everything in **Controls** is more convenient and in line with what is in the main WPF, we can consider it that way.
> 
> Yep

Well, then, just like the main WPF, we can omit decorative subfolders and keep everything in one place, taking care of the correctness of the namespace.
https://github.com/dotnet/wpf/tree/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls

IvanDmitriev1:
Then I prefer the directory tree.

IvanDmitriev1:
But at the same time winui 3.
https://github.com/microsoft/microsoft-ui-xaml/tree/main/dev

pomianowski:
> 

Well yes, but it's a CPP project

IvanDmitriev1:
In my opinion , c++ namespace structured by folders, at least I did.

pomianowski (approved):

PR #586, opened 2023-04-10
Templates: Fix async code and nullability issues

rodion-m:
- Fix async code issues in App.xaml.cs and ApplicationHostService.cs
- Make the method `App.GetService` nullable

## Pull request type

- [ ] Bugfix

Incorrect calling of async methods in the `App.OnStartup` and `App.OnExit` methods, cause `void` is not awaitable.

Issue Number: N/A

## What is the new behavior?
This PR makes the `App.OnStartup` and `App.OnExit` methods synchronous so that WPF waits for them to be executed.


pomianowski:
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#avoid-using-taskresult-and-taskwait

PR #602, opened 2023-04-23
Update `ToggleSwitch` add `OffContent` and  `OnContent`

zh3305:
## Pull request type

- [x] Update
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

动画


LHTXQ:
This change decision is unreasonable, content should be driven by binding variables instead of setting content separately for both states.

zh3305:
> This change decision is unreasonable, content should be driven by binding variables instead of setting content separately for both states.
I don't want to bind too many variables to handle this switch change, and I don't want to handle this unique change by converting when there are multiple states such as switch, split, and close, which makes the code bloated. I looked at the WinUI 3 Gallery code and it solved my problem perfectly. With this change you can still use the Content property to set the label, OffContent OnContent is just an option

动画

LHTXQ:
> > This change decision is unreasonable, content should be driven by binding variables instead of setting content separately for both states.
> 
> 
> I don't want to bind too many variables to handle this switch change, and I don't want to handle this unique change by converting when there are multiple states such as switch, split, and close, which makes the code bloated. I looked at the WinUI 3 Gallery code and it solved my problem perfectly. With this change you can still use the Content property to set the label, OffContent OnContent is just an option
> 
> 动画
> 
> 

🤔行吧……

src/Wpf.Ui/Controls/ToggleSwitch.cs
```diff
@@ -15,4 +16,23 @@ namespace Wpf.Ui.Controls;
 [ToolboxBitmap(typeof(ToggleSwitch), "ToggleSwitch.bmp")]
 public class ToggleSwitch : System.Windows.Controls.Primitives.ToggleButton
 {
+    public static readonly DependencyProperty OffContentProperty = DependencyProperty.Register(
+        "OffContent", typeof(object), typeof(ToggleSwitch), new PropertyMetadata(null));
+
+    [Bindable(true)]
+    public object OffContent
```
pomianowski, line 23: Missing docs

pomianowski (approved):

pomianowski (approved):

pomianowski (approved):

pomianowski (approved):

PR #627, opened 2023-05-06
Let TitleBar show 'SystemMenu' on mouse right button up.

LHTXQ:
--- Commit by LHTXQ

Show 'SystemMenu' on mouse right button up.

## Pull request type

- [ ] Update
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Do nothing on mouse right button up.

Issue Number: N/A

## What is the new behavior?

Show 'SystemMenu' on mouse right button up.

image


LHTXQ:
This is my first pull request, it seems something wrong in the progress, please tell me how to make it right.

LHTXQ:
Maybe we can give the SystemMenua a more modern appearance in keeping with the Wpfui style，but I don't know how to modify it.

In addition, we can also add a function to let user insert or delete menu items in SystemMenu(just like Edge browser's SystemMenu).

chucker:
>Maybe we should give the SystemMenua a more modern appearance in keeping with the Wpfui style，but I don't know how to modify it.

Perhaps it would be easier to replace it with a custom menu altogether. Then we could give people a high-level WPF API where they can simply add additional items, and perhaps boolean properties to remove some default items.

oli-g-sk:
> Maybe we should give the SystemMenua a more modern appearance in keeping with the Wpfui style，but I don't know how to modify it.

I think even Windows isn't sure about how it's supposed to look like yet. I just checked Notepad (which uses the Fluent look for some time), and the menu looks identical to your screenshot.

image

Then I checked the Settings app, and almost looks the same, but it respects the dark theme.

image
 
> In addition, we can also add a function to let user insert or delete menu items in SystemMenu(just like Edge browser's SystemMenu).

IMO that would be nice, however I would definitely think of it as a separate feature to follow this one.

With this, we're just providing a way to invoke native functionality that was not available (compared to WPF). Something that should just work "out of the box" and doesn't require any "design decisions" as to how to expose it to the users of the library.

pomianowski (approved):

Issue #635, opened 2023-05-12
The FontSize Stlye


MakesYT:
### Describe the bug

The FontSize property of TextBox has no effect on the size of the Icon
And the ClearButton display is abnormal. It should be centered, but in fact it is up (on 2.0.3)/ not displayed (on 3.0.0pre2).
`
<ui:TextBox ClearButtonEnabled="True" FontSize="32">
            <ui:TextBox.Icon>
                <ui:SymbolIcon   Symbol="Search48"  FontSize="32"/>
            </ui:TextBox.Icon>
        </ui:TextBox>
`

### To Reproduce

1.  No special steps, only install from Nuget
2. `
<ui:TextBox ClearButtonEnabled="True" FontSize="32">
            <ui:TextBox.Icon>
                <ui:SymbolIcon   Symbol="Search48"  FontSize="32"/>
            </ui:TextBox.Icon>
        </ui:TextBox>
`

### Expected behavior

The size of the Icon should be controlled by FontSize or some other property,
And the ClearButton display at the center.

### Screenshots

![image](https://github.com/lepoco/wpfui/assets/42534870/5d4693a8-1199-49fe-bb54-99e1715cc65e)

### OS version

版本	Windows 11 专业工作站版
版本	22H2
安装日期	‎2022/‎7/‎27
操作系统版本	22624.1690
体验	Windows Feature Experience Pack 1000.22641.1000.0

### .NET version

net7.0-windows

### WPF-UI NuGet version

3.0.0pre2

bbsuuo:
我猜你布局有问题

MakesYT:
> 我猜你布局有问题

也许反正改不了大小

MakesYT:
> 我猜你布局有问题

也许反正改不了大小

niels9001:
Fixed in #638

PR #637, opened 2023-05-14
Fix Controls bug

MakesYT:
…width of the TextBoxClearButton

## Pull request type

- [ ] Update
- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #635 

## What is the new behavior?

1. Fix LeftFluentNavigationViewItemTemplate content fontsize binding
image
image

2. Fix TextBox Icon size,Add two properties to define the icon size and width of the TextBoxClearButton
    Also make icon and the TextBoxClearButton at the center
 image
 image
 image
 It should be like this
 image
 but it is like this
  image
 It only works when I hot reload the binding, I don't know how to fix this

## Other information

- There are some bugs that need help


MakesYT:
Sorry committed to the wrong branch

PR #754, opened 2023-09-09
Updated the NumberBox ValueProperty's default BindingMode to be TwoWay

syntax-tm:
## Pull request type

- [x] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, a `Binding` to the `NumberBox.Value` is `OneWay` by default. The `Source` (`DataContext`) raising an `INotifyPropertyChanged` event will trigger a change in the `NumberBox`. However, a change in the `NumberBox` will not update the `Source`.

If you are binding to the `NumberBox.Value` (using the default `BindingMode`, `OneWay`):

- The increment, decrement, and clear operations are not updating the `Source`
  - This is made worse by the fact the `Text` displayed will not match the actual `Value`
  - The `Value` property of the `NumberBox` changing means that will no longer match the `Source` property
- Manually inputting (typing) a number does not update the `Source` property

I created a test project to demonstrate this, but here's some screenshots that show the issue that currently exists (OneWay) and then the solution (TwoWay).

<table>
  <tr>
    <td>
      <p align="center">
        
      </p>
    </td>
    <td>
      <p align="center">
        
      </p>
    </td>
  </tr>
</table>

[WpfDemoApp.zip](https://github.com/lepoco/wpfui/files/12566857/WpfDemoApp.zip)

## What is the new behavior?

- `NumberBox.Value` default `BindingMode` is now `TwoWay` and functions as expected

## Other information

I was surprised that this wasn't already set for this control and I'm not sure if there's any reasoning for why that is but here's an excerpt from the MSDN docs on [BindingMode](https://learn.microsoft.com/en-us/dotnet/api/system.windows.data.bindingmode?view=windowsdesktop-7.0) regarding the default bindings of "_user-editable control properties_":

> _The default is [Default](https://learn.microsoft.com/en-us/dotnet/api/system.windows.data.bindingmode?view=windowsdesktop-7.0#system-windows-data-bindingmode-default), which returns the default binding mode value of the target dependency property. However, the default value varies for each dependency property. **In general, user-editable control properties, such as those of text boxes and check boxes, default to two-way bindings, whereas most other properties default to one-way bindings.**_ ([source](https://learn.microsoft.com/en-us/dotnet/api/system.windows.data.binding.mode?view=windowsdesktop-7.0))


syntax-tm:
I'm not sure what method of contact you guys prefer for the signing the Contributor License Agreement, but if it's possible you can consider this my acceptance of the agreement. If there's anything more you need from me just let me know. Thanks!

pomianowski (approved):

PR #759, opened 2023-09-12
Fix submenu navigation in NavigationViewItem

koal44:
The feature for submenus in NavigationViewTop was previously implemented but left unused prolly because of odd behavior such as stack overflow errors in AutoSuggest.

The problem was traced to the NavigationViewItem.MenuItems dependency property, where a new ObservableCollection was incorrectly used as the default value. According to [MS Docs](https://learn.microsoft.com/en-us/dotnet/desktop/wpf/properties/collection-type-dependency-properties?view=netdesktop-7.0), "dependency property metadata shouldn't include a default reference-type value because that value will be assigned to all instances of the class, creating a singleton class."

## Pull request type

- [ ] Update
- [x] Bugfix
- [x] Feature
- [ ] Build related changes

In the Gallery app if you add submenu items to NavigationView.Top you will get stack overflow
errors on AddItemsToAutoSuggestBoxItems(IList list) for menuItems.

  <ui:NavigationView PaneDisplayMode="Top">
    <ui:NavigationView.MenuItems>
      <ui:NavigationViewItem  Content="Menu Item 1"
        <ui:NavigationViewItem.MenuItems>
          <ui:NavigationViewItem Content="Menu SubItem 1" />
          <ui:NavigationViewItem Content="Menu SubItem 2" />
        </ui:NavigationViewItem.MenuItems>
      </ui:NavigationViewItem>

Issue Number: N/A

## What is the new behavior?

- Change MenuItems to an ObservableCollection<object> and initialize in ctor
- Fix related properties such as HasMenuItems and MenuItemsSource (removed)
- Add demo in the Gallery app

## Other information

image

### DIFF
--- a/Settings.XamlStyler
+++ b/Settings.XamlStyler
@@ -7,7 +7,7 @@
     "SeparateByGroups": false,
     "AttributeIndentation": 0,
     "AttributeIndentationStyle": 1,
-    "RemoveDesignTimeReferences":  false,
+    "RemoveDesignTimeReferences": false,
     "IgnoreDesignTimeReferencePrefix": false,
     "EnableAttributeReordering": true,
     "AttributeOrderingRuleGroups": [
@@ -38,5 +38,7 @@
     "ThicknessSeparator": 2,
     "ThicknessAttributes": "Margin, Padding, BorderThickness, ThumbnailClipMargin",
     "FormatOnSave": true,
-    "CommentPadding": 2
+    "CommentPadding": 2,
+    "IndentSize": 4,
+    "IndentWithTabs": false
 }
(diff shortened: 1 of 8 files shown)


syntax-tm:
Why did you remove the `MenuItemsSource`?

`Items` is for defining in XAML. This is usually an `object[]` but it contains `FrameworkElements`.
`ItemsSource` is for databinding to a collection. Typically uses a `DataTemplate` to display the items.

- [ItemsControl.Items Property](https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.itemscontrol.items)
- [ItemsControl.ItemsSource Property](https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.itemscontrol.itemssource)

If one of those was broken, then it's better to fix it instead of removing one of the options. You should be able to define a menu directly in XAML or bind to an `IEnumerable` source (with support for `INotifyCollectionChanged`).

koal44:
Here's an [update ](https://github.com/koal44/wpfui/tree/menuitems-vm-test) on `MenuItems` working with both xaml and viewmodels. I'm still trying to understand why MS uses a source property for `ItemsControl`; it seems unneeded here. But i prolly lack the experience to make that call.

image

PR #762, opened 2023-09-14
Implement PassiveScrollViewer; replace ScrollViewer in all XAML controls

koal44:
Implement `PassiveScrollViewer` across various controls for improved mousewheel behavior

## Pull request type

- [x] Feature

## Current Behavior
Currently, many controls in this library use a `ScrollViewer`, which, while configurable (`CanContentScroll`, etc) still consumes mousewheel events, leading to a less-than-optimal user experience [Stackoverflow](https://stackoverflow.com/questions/76131100/how-do-i-prevent-a-scrollviewer-in-wpf-from-eating-the-scroll-event-if-the-scrol). This can be observed within the Gallery app by going to the 'NavigationView' section and attempting to mouse wheel down the page. While replacing .NET's control might seem heavy-handed the only current remedy for app devs is to copy the control and directly replace the style.

## New Behavior
1. Open the Gallery app and go to the 'NavigationView' page
2. Use the mouse wheel to scroll through the controls down the center of the page (it's cheating if you mouse wheel over the scrollbar).
3. Optionally, toggle the `IsScrollSpillEnabled` default value `new PropertyMetadata(true));` (PassiveScrollViewer.cs, line 21) and head over to the 'DataGrid' page to observe the difference.


pomianowski (approved):

pomianowski:
Hey @koal44, thank you for your contributions to the WPF UI. It seems that event bubbling works perfectly and scroll correctly moves to higher controls.

Issue #593, opened 2023-04-16
Unhandled File not found 0x80070002 in DwmGetColorizationParameters


wfjsw:
### Describe the bug

The DWM api is throwing a `System.IO.FileNotFoundException` when accessing `Dwmapi.DwmGetColorizationParameters(out DWMCOLORIZATIONPARAMS dwParameters)`

### To Reproduce

Unknown

### Expected behavior

The error is caught and a default color is applied

### OS version

Windows 10.0.22621

### .NET version

.NET 6.0.16

### WPF-UI NuGet version

https://github.com/lepoco/wpfui/commit/91a708991a69f168943e8d961cc123bf8ca92965

### Additional context

```
System.IO.FileNotFoundException: 系统找不到指定的文件。 (0x80070002)
  ?, in void Dwmapi.DwmGetColorizationParameters(out DWMCOLORIZATIONPARAMS dwParameters)
  ?, in Color UnsafeNativeMethods.GetDwmColor()
  ?, in void Theme.Apply(ThemeType themeType, WindowBackdropType backgroundEffect, bool updateAccent, bool forceBackground)
  ?, in void Watcher.UpdateThemes(SystemThemeType systemTheme)
  ?, in void Watcher.Watch(Window window, WindowBackdropType backgroundEffect, bool updateAccents, bool forceBackground)+(object sender, RoutedEventArgs args) => { }
  ?, in void RoutedEventHandlerInfo.InvokeHandler(object target, RoutedEventArgs routedEventArgs)
  ?, in void EventRoute.InvokeHandlersImpl(object source, RoutedEventArgs args, bool reRaised)
  ?, in void UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
  ?, in void UIElement.RaiseEvent(RoutedEventArgs e)
  ?, in void BroadcastEventHelper.BroadcastEvent(DependencyObject root, RoutedEvent routedEvent)
  ?, in object BroadcastEventHelper.BroadcastLoadedEvent(object root)
  ?, in void MediaContext.FireLoadedPendingCallbacks()
  ?, in void MediaContext.FireInvokeOnRenderCallbacks()
  ?, in void MediaContext.RenderMessageHandlerCore(object resizedCompositionTarget)
  ?, in object MediaContext.RenderMessageHandler(object resizedCompositionTarget)
  ?, in void MediaContext.Resize(ICompositionTarget resizedCompositionTarget)
  ?, in void HwndTarget.OnResize()
[...]

PR #859, opened 2023-12-06
DwmGetColorizationParameters fallback in case of failure

davidegiacometti:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/lepoco/wpfui/issues/593

## What is the new behavior?

- Fallback to registry, then to Windows default accent color in case of `DwmGetColorizationParameters` failure .
- I have tested it on Windows 10 22H2: I am able to reproduce the issue sometimes while turning on/off high contrast or switching between high contrast modes.
- I haven't been able to test it on Windows 11 but I have verified that the registry key is present and contains the expected value.


stefansjfw:
@pomianowski Does the latest version of the package include this fix? Thanks

Issue #872, opened 2023-12-11
ScrollBar buttons are not aligned


niels9001:
### Describe the bug

The up/down icons are not center aligned

### To Reproduce

Use a scrollbar

### Expected behavior

Center aligned

### Screenshots

![image](https://github.com/lepoco/wpfui/assets/9866362/6433fcab-78cf-4f69-a42b-216a894be92a)

### OS version

W11

### .NET version

NET 8

### WPF-UI NuGet version

Preview-12

Issue #871, opened 2023-12-11
Bug title


niels9001:
### Describe the bug

The scroller in `ListView` should render on top of the `ListViewItems`, instead of next to it. Resulting in: https://github.com/microsoft/PowerToys/issues/30318

WPFUI:
![image](https://github.com/lepoco/wpfui/assets/9866362/9e29457d-5df2-4586-863b-ea960bffbde2)

### To Reproduce

Make a `ListView` stretch the entire width of the window

### Expected behavior

The scroller should render on top of the items, not next to it:

![image](https://github.com/lepoco/wpfui/assets/9866362/6cfa0d0b-9e4d-47bf-822b-ceb58fe83729)

### OS version

W11

### .NET version

NET 8

### WPF-UI NuGet version

Preview-12

htcfreek:
@pomianowski 
Is the fix PR included in preview 13?

PR #873, opened 2023-12-11
Scrollbar fixes

niels9001:
### Updated the (Dynamic)ScrollBar default style to be more inline with WinUI addressing: #872
- Added on hover background color (couldn't use `InAppAcrylicDefaultBrush` like in WinUI, but used the replacement color we also use for e.g. `Flyout`)
- Decreased scrollbar thumb width based on WinUI spec
- Decreased icon fontsize so it's centered and inline with WinUI spec

Before:
image

After:
image

### ScrollBar now overlaps content, addressing: #871
- Tweaked the default `ScrollViewer` style so it overlaps its content
- Set a default style for `PassiveScrollViewer` so it uses the `ScrollViewer` style as well (as `ListView` uses `PassiveScrollViewer`)
Before:
image

After:
image


niels9001:
@pomianowski In case this PR gets approved, would you mind creating a `Preview.13` :)?

pomianowski (approved):

PR #931, opened 2024-02-05
Fix focus issues with ContentDialog

jackBonadies:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, ContentDialog has 2 focus issues (that can be observed in the wpfui gallery).  1. On launching ContentDialog focus is not on content dialog, it stays on the background controls.  2. One can tab off the ContentDialog controls onto background controls.  

Issue Number: N/A

## What is the new behavior?

It now behaves like WinUI 3 - i.e. it has focus starting on the ContentDialog and one cannot tab off ContentDialog (it cycles within the dialog).  Note that there was an attempt to Focus() dialog on load, but since ContentDialog was not set to focusable that call returned false (and did nothing).  This is now fixed.

## Other information

This is just a candidate solution.  If one does not want ContentDialog set to Focusable for whatever reason, then they can write a routine to select the first focusable child.


pomianowski:
It looks almost too elegantly simple, thanks for your support @jackBonadies

pomianowski (approved):

PR #979, opened 2024-02-28
Fix ListView to be compatible when ListView.View is a GridView

koal44:
The UI library currently breaks all `ListView` controls that use the `GridView` view. This PR adds support for `GridView` and showcases it in the GalleryApp
## Pull request type

- [x] Bugfix
- [x] Feature

The current behavior is that UI library interferes with how different derived types of `BaseView` are dynamically stylized, which renders types like `GridView` incompatible with this library. Although you can workaround this by voiding the library's styles, something like:
```
<ListView Style="{x:Null}">
   <ListView.Resources>
      <Style TargetType="ListViewItem"> <Setter Property="Background" Value="Transparent" /> </Style>
   </ListView.Resources>
   <ListView.View>
      <GridView><GridViewColumn DisplayMemberBinding="{Binding ...}" /></GridView>
   </ListView.View>
</ListView>
  ```
Digging through older framework code, it appears that when `View` changes, `ListView` uses the current `View` to change its own `DefaultStyleKey` and also apply new keys to its viewitem children. When the `View` is null, the `DefaultStyleKey` will point to `ListBox` and `ListBoxItem` (theme styles) and for `GridView` it'll be `GridViewStyle` and `GridViewItemContainerStyle`. When this library directly set a style on `ListView` and `ListViewItem` it broke the unorthodox approach the framework used to dynamically set styles to different views.

My first attempt was to go with the flow of the WPF framework. For example, you should be able to style `ListBox` and see that refelcted in a `ListView` with null `View`, but various attempts at this failed.

Since there is actually only one derived type of BaseView - namely `GridView`, there are effectively only two scenarios that exist - nullview and gridview, so we can use a more standard approach and use xaml styles to handle these two different cases.

## What is the new behavior?

- Extend ListView so that we can subscribe to changes in `View`
- Use triggers in `ListView.xaml` and `ListViewItem.xaml` to adapt to different View states
- Add a sample to the the Gallery app

## Other information
image

### DIFF
--- a/src/Wpf.Ui/Controls/ListView/ListView.cs
+++ b/src/Wpf.Ui/Controls/ListView/ListView.cs
@@ -1,0 +1,60 @@
+namespace Wpf.Ui.Controls;
+
+public class ListView : System.Windows.Controls.ListView
+{
+    public string ViewState
+    {
+        get => (string)GetValue(ViewStateProperty);
+        set => SetValue(ViewStateProperty, value);
+    }
+
+    /// <summary>Identifies the <see cref="ViewState"/> dependency property.</summary>
+    public static readonly DependencyProperty ViewStateProperty = DependencyProperty.Register(nameof(ViewState), typeof(string), typeof(ListView), new FrameworkPropertyMetadata("Default", OnViewStateChanged));
+
+    private static void OnViewStateChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+    {
+        if (d is not ListView self)
+        {
+            return;
+        }
+
+        self.OnViewStateChanged(e);
+    }
+
+    protected virtual void OnViewStateChanged(DependencyPropertyChangedEventArgs e)
+    {
+        // derived classes can hook `ViewState` property changes by overriding this method
+    }
+
+    public ListView()
+    {
+        Loaded += OnLoaded;
+    }
+
+    private void OnLoaded(object sender, RoutedEventArgs e)
+    {
+        // immediately unsubscribe to prevent memory leaks
+        Loaded -= OnLoaded;
+
+        // get the descriptor for the `View` property since the framework doesn't provide a public hook for it
+        var descriptor = DependencyPropertyDescriptor.FromProperty(System.Windows.Controls.ListView.ViewProperty, typeof(System.Windows.Controls.ListView));
+        descriptor?.AddValueChanged(this, OnViewPropertyChanged);
+        UpdateViewState(); // set the initial state
+    }
+
+    private void OnViewPropertyChanged(object? sender, EventArgs e)
+    {
+        UpdateViewState();
+    }
+
+    private void UpdateViewState()
+    {
+        var viewState = View is null ? "Default" : "GridView";
+        SetCurrentValue(ViewStateProperty, viewState);
+    }
+
+    static ListView()
+    {
+        DefaultStyleKeyProperty.OverrideMetadata(typeof(ListView), new FrameworkPropertyMetadata(typeof(ListView)));
+    }
+}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by koal44** [comment]
Just wanted to add there seems to be a few open issues relevant to this bugfix: #918 #436 #198

File: src/Wpf.Ui/Controls/ListView/ListView.cs [hunk 1]
```diff
@@ -0,0 +1,60 @@
+namespace Wpf.Ui.Controls;
+
+public class ListView : System.Windows.Controls.ListView
```
pomianowski, line 3: Missing docs and examples for public control, see: 
https://github.com/lepoco/wpfui/blob/baaf749864df48be57c18821c9dcd47a2b5faa73/src/Wpf.Ui/Controls/Button/Button.cs#L12-L32

pomianowski, line 3: New control should be added to https://github.com/lepoco/wpfui/blob/baaf749864df48be57c18821c9dcd47a2b5faa73/src/Wpf.Ui/VisualStudioToolsManifest.xml

src/Wpf.Ui/Controls/ListView/ListView.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    public string ViewState
```
pomianowski, line 5: Missing docs for pubplic property

pomianowski (changes requested):

pomianowski:
Hey @koal44 , thanks for your contribution to the project. I would add some documentation and improve the formatting

koal44:
I force pushed the last commit so that it aligned better with Wpf.Ui/development.

The latest commit removes the custom `ui:ListView` class that was previously introduced in this PR.  The purpose of the class was to expose the `ListView`'s view state (either "Default" or "GridView"`) but XAML can bind just as easily to the standard `ListView.View` property with a `ValueConverter`.

I didn't mention it before, but the XAML style for `GridView`  was adapted from the Aero2 theme with slight changes to fit this project's aesthetics. The `GridViewColumnHeader`'s padding and margins were eyeballed to fit with the GalleryApp and should probably be reviewed so that alignment is consistent between header columns and data columns for all user apps.

src/Wpf.Ui/Controls/ListView/ListViewViewState.cs
```diff
@@ -0,0 +1,12 @@
+// This Source Code Form is subject to the terms of the MIT License.
+// If a copy of the MIT was not distributed with this file, You can obtain one at https://opensource.org/licenses/MIT.
+// Copyright (C) Leszek Pomianowski and WPF UI Contributors.
+// All Rights Reserved.
+
+namespace Wpf.Ui.Controls;
+
+public enum ListViewViewState
```
pomianowski, line 8: Missing docs for public api

koal44:
Ok, the recent update should address reviewer concerns about documentation and VS tooling. I also re-implemented the custom `ui:ListView` class as it should leave the original framework's `ListView` unchanged and unbugged.

pomianowski (approved):

david-risney:
Looking forward to these ListView PRs showing up in a release! This is great, thank you!

pomianowski:
> Looking forward to these ListView PRs showing up in a release! This is great, thank you!


Hey, already released in the version 3.0.3

BingGitCn:
Tried setting the IsEnabled property to false?

david-risney:
> > Looking forward to these ListView PRs showing up in a release! This is great, thank you!
> 
> Hey, already released in the version 3.0.3

I'm trying out the latest version and its looking great! Thanks so much!

PR #993, opened 2024-03-13
Fix Close Button icon color in TitleBar while mouse over

seasonyuu:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Normally, the TitleBar should change Close button background to Red, and change foreground to White while mouse over. But it looks like this now.

image

Issue Number: N/A

## What is the new behavior?

image

Because the `TitleBar.cs` hook mouse event, so the TemplateBinding in `TitleBar.xaml` cannot trigger by `IsMouseOver`. So this pull request continuely use the `Hover/RemoveHover` method in `TitleBarButton.cs` as event, and add a `MouseOverButtonsForeground` property to allow user defining this color.

-
-

### DIFF
--- a/src/Wpf.Ui/Controls/TitleBar/TitleBar.xaml
+++ b/src/Wpf.Ui/Controls/TitleBar/TitleBar.xaml
@@ -15,6 +15,7 @@
         <Setter Property="Height" Value="30" />
         <Setter Property="MouseOverBackground" Value="{Binding Path=ButtonsBackground, RelativeSource={RelativeSource AncestorType={x:Type controls:TitleBar}}}" />
         <Setter Property="ButtonsForeground" Value="{Binding Path=ButtonsForeground, RelativeSource={RelativeSource AncestorType={x:Type controls:TitleBar}}}" />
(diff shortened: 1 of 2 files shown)


pomianowski:
hey @seasonyuu, thanks for your contribution to the project

koal44:
This PR has some bugs that should be worked out.
1) Binding failures in TitleBar.xaml
```
<Setter Property="MouseOverButtonsForeground" Value="{Binding Path=MouseOverButtonsForeground, RelativeSource={RelativeSource AncestorType={x:Type controls:TitleBar}}}" />
```
There's no such property for `TitleBar` but there was one introduced for `TitleBarButton`. But anyhoo, the line can be omitted as the styling logic is entirely handled in code through `TitleBarButton.Hover()`

2) The `ButtonsForeground` is now broken when switching themes
To reproduce the bug: run the gallery app, hover over the TitleBar buttons (like minimize or close buttons), then switch themes and you'll see that the foreground is now stuck to the previous theme.
The reason for this is `ButtonsForeground = _cacheButtonsForeground`, which breaks the theme binding from
`<Setter Property="ButtonsForeground" Value="{DynamicResource ButtonForeground}" />`

koal44:
image
as seen here, the buttons become invisible when switching to light theme

seasonyuu:
Got it. Give me sometime

koal44 ***@***.***> 于 2024年3月17日周日 04:52写道：

PR #1022, opened 2024-03-27
feat(controls): Add `NumericUpDown` control

### LINKED ISSUE #1673: NumberBox's SpinButtonPlacementMode="Compact" behaves like "Hidden"
### Describe the bug

When I set the attribute to Compact, I expect the spin buttons to behave like WinUI's and as described here [wpfui.lepo.co/api/Wpf.Ui.Controls.NumberBoxSpinButtonPlacementMode.html](https://wpfui.lepo.co/api/Wpf.Ui.Controls.NumberBoxSpinButtonPlacementMode.html) (i.e. horizontal placement when box isn't focused)

But they behave like Hidden (no spin button at all)

### To Reproduce

Set the attribute and run the app.

### Expected behavior

They behave as described

### OS version

11

### .NET version

10

### WPF-UI NuGet version

4.2.0

### Additional context

it seems https://github.com/lepoco/wpfui/pull/1022 fixes/implements the behaviour (although as a fully distinct new component/control, which isn't ideal)

koal44:
## Pull request type

- [x] Feature

## What is the new behavior?

Added new control `NumericUpDown`. The library already has a `NumberBox` control, but this is my version.

numeric-up-down

- Implemented new control `NumericUpDown`
- Showcased it in the Gallery App


pomianowski:
Hey @koal44, thanks for putting the work into the control, but we should avoid creating two controls to do the same thing. If you have any ideas on how to improve `NumberBox`, I think this would be a better approach.

LucaSamarCMA:
it would be great to have this control

chucker:
I don't see why we need two controls that do the same thing.

But I _can_ see simplifying this PR to introduce a new `NumberBoxSpinButtonPlacementMode.Classic`, which looks like the screenshot above: the up/down buttons are vertically arranged.

stanlrt:
Shouldnt that PR be transformed into a fix for https://github.com/lepoco/wpfui/issues/1673 ? NumberBox is already supposed to support what you implemented, but it's broken

stanlrt:
> I don't see why we need two controls that do the same thing.
>
> But I *can* see simplifying this PR to introduce a new `NumberBoxSpinButtonPlacementMode.Classic`, which looks like the screenshot above: the up/down buttons are vertically arranged.

It was my understanding this is precisely what `NumberBoxSpinButtonPlacementMode.Compact`  is already supposed to do, but maybe I read docs wrong? In WinUI that is for sure what it does though.

chucker:
> > I don't see why we need two controls that do the same thing.
> > But I _can_ see simplifying this PR to introduce a new `NumberBoxSpinButtonPlacementMode.Classic`, which looks like the screenshot above: the up/down buttons are vertically arranged.
> 
> It was my understanding this is precisely what `NumberBoxSpinButtonPlacementMode.Compact` is already supposed to do, but maybe I read docs wrong? In WinUI that is for sure what it does though.

I think that's correct.

chucker:
I dug deeper, and `Classic` and `Compact` aren't the same. I've implemented `Compact` in #1709, roughly the way WinUI does it.

- both Classic and Compact display a vertical arrangement of up/down buttons
- Compact, when clicking on those, will show a Flyover with larger, now properly separated buttons. _Those_ buttons then actually increment/decrement the value

image

image

- Classic, OTOH, lets you increment/decrement _with_ those tiny arrow buttons

image

I imagine the WinUI team opted against offering this option, since it's basically impossible to use in a touch environment.

pomianowski:
Continued by #1709

Issue #1025, opened 2024-03-28
TitleBar suggestion


bunnyi116:
### Is your feature request related to a problem? Please describe

Within the `TitleBar`, there is a `Header`, but I prefer not to use it as it aligns to the right only. Therefore, I would like to further break down these elements into smaller controls to enhance the flexibility of custom windows.

### Describe the solution you'd like

I think dividing the `TitleBar` control into three separate controls would be beneficial: `WindowTitle`, `WindowBar`, and `WindowButton`.

- `WindowTitle`: Responsible for displaying the window title.
- `WindowBar`: Contains basic functionalities such as dragging.
- `WindowButton`: Includes maximize, minimize, and close buttons (and potentially further subdivided into smaller controls).

I believe this approach can provide more customization options for windows, freeing them from the constraints of the `TitleBar`.

![image](https://github.com/lepoco/wpfui/assets/37466008/661a66fb-251f-4ea5-9375-65aad740cc92)

Difegue:
Been hitting this issue as well; Although I believe adding a separate `LeftHeader` property e.g https://github.com/lepoco/wpfui/issues/960 would be better.

luca-domenichini:
see also #1483 for allowing centered content too

Issue #775, opened 2023-09-29
TitleBar no longer offers way to replace Title with complex content


chucker:
### Is your feature request related to a problem? Please describe

In the 3.0 previews, `TitleBar`'s `Header` property has been changed in behavior. It now shows up _to the right of_ the `Title`, whereas it used to be _instead of_ it.

I would appreciate if we could get the old behavior back, perhaps as a separate property `LeftHeader`.

### Describe the solution you'd like

It depends on what this new behavior is even good for; the gallery doesn't contain a sample.

I would be fine with a set of `LeftHeader` and `RightHeader` properties.

### Additional context

I use this to enable macOS-like two-line titles, e.g.

<img width="1456" alt="image" src="https://github.com/lepoco/wpfui/assets/20194/e7a07c4e-487d-4a53-8a11-ad06ae584a2c">

jonlipsky:
I would love to see this as well.  I want to create a titlebar similar to Visual Studio Code:

<img width="444" alt="image" src="https://github.com/lepoco/wpfui/assets/1466386/fd32e448-f856-416d-a790-02048fa48e63">

PR #1030, opened 2024-03-31
Add support for setting content in place of the title in a TitleBar

jonlipsky:
## Pull request type

- [ ] Update
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently there is no way to set content in a `TitleBar` that is left justified and/or use content other than the default `TextBlock` for the title content. 

Issue Number: [#775] [#1025]

## What is the new behavior?

This PR adds support for (optionally) specifying content to be displayed as a replacement for the `TextBlock` which displays the title.  The following changes have been made:

- Added a `LeftHeader` property that can be optionally set to specify the view to be presented in place of the default `TextBlock` that displays the title (when set).
- A `TextBlock` is initialized in the `TitleBar` constructor and assigned to the `LeftHeader` property so that setting the title is still the default behavior.
- Renamed `Header` to `RightHeader` to make the API more self-documenting as to the purpose of the property.
- Updated the sandbox window example to demonstrate this new functionality.

image

## Other information

- This change is a breaking change from the previous version since `Header` has been renamed to `RightHeader`.  One remedy would be to add a new property called `Header`, which would delegate to `RightHeader` and be marked as deprecated to give users time to update their code.

- One additional enhancement that may be nice to have would be the ability to vertically align the Minimize, Maximize and Close buttons of the window.  It may be more aesthetically pleasing to have them center aligned depending on the content in the RightHeader.  As an example, JetBrains Rider does this in their custom titlebar:

image


537mfb:
To prevent backward compatibility issues you could have kept Header and just have added LeftHeader.

Once there's a major version change that can break backwards compatibility then a name change from Header to RightHeader could be considered without issue

jonlipsky:
@537mfb Yes, I'd be happy to submit an update to the pull request to change the naming of `RightHeader` back to `Header`.  I renamed that property because the original behavior was not obvious (in my personal opinion).

chucker:
Given that `Header`'s behavior changed between 2.x and 3.x, and this would change it back, I would argue the rename makes sense. Giving it a new name makes it clear that the behavior has changed.

Jay-o-Way:
I think you guys are forgetting that "left" and "right" _should_ be swapped when the OS is using a right-to-left language. Better use something like "primary" and "secondary, or something like that.

Jay-o-Way:
> * the ability to vertically align the Minimize, Maximize and Close buttons

That would be setting the **PreferredHeightOption** to **TitleBarHeightOption.Tall**
https://learn.microsoft.com/windows/apps/develop/title-bar#tall-title-bar-support-for-custom-title-bars

src/Wpf.Ui/Controls/TitleBar/TitleBar.xaml
```diff
@@ -118,42 +118,41 @@
...
+                                <ColumnDefinition Width="Auto" />
                                 <ColumnDefinition Width="*" />
                                 <ColumnDefinition Width="Auto" />
                             </Grid.ColumnDefinitions>
 
-                            <!--  Additional header content  -->
+                            <!--  Title text or other header content  -->
                             <ContentPresenter
                                 Grid.Column="0"
+                                HorizontalAlignment="Left"
```
Jay-o-Way, line 144: No need for this property when width=auto

jonlipsky:
@Jay-o-Way I initially called the renamed properties `LeadingHeader` and `TrailingHeader`; however it wasn't clear to me that the titlebar class was adaptive for right-to-left languages the way it is built, thus I renamed them.   I agree that if it is adaptive, then they should have an alternate name.

537mfb:
Not sure how easy this would be to achieve but how about having Header on the side of the system buttons (as is now) and make Title work like Content in other controls where if you give a string it works like a textblock but you can give it other arbitrary content.

Think of a Label - you can give it a string or a layout with several controls as content

jonlipsky:
@537mfb That's essentially what my PR does right now.  By default, it's a `TextBlock`, however if you set `LeftHeader`, that content becomes whatever you set it to.

537mfb:
@jonlipsky my idea is slightly different.

you could have:

`<iu:TitleBar Title="My App Name" ...`

and it would work as it does now. But you could also do:

```
<iu:TitleBar ....>
    <iu:TitleBar.Title>
         <StackPanel Orientation="Horizontal">
             <ui:SymbolIcon Symbol="Fluent24"/>
             ....
         </StackPanel>
    </iu:TitleBar.Title>
</iu:TitleBar>
```

pomianowski:
To be honest, I'm not a fan of the division into HeaderLeft and HeaderRight. I would prefer one big Header as a whole Titlebar up to navigation buttons with default value - Icon and Application Name

Summpot:
Any update?

chucker:
> To be honest, I'm not a fan of the division into HeaderLeft and HeaderRight. I would prefer one big Header as a whole Titlebar up to navigation buttons with default value - Icon and Application Name, similar to @537mfb's suggestion

[MAUI apparently now uses](https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview7/dotnetmaui.md#new-titlebar-control-and-windowtitlebar-for-windows) the notion of `TrailingContent`, which is roughly what `HeaderRight` is in this PR:

jonlipsky:
If a project admin/leader lets me know their preference as to how they want it to work, I'd happily amend the PR to implement that.  I took the initial approach I did because there were other issues/requests asking for it that way.

SamKr:
> If a project admin/leader lets me know their preference as to how they want it to work, I'd happily amend the PR to implement that. I took the initial approach I did because there were other issues/requests asking for it that way.

@pomianowski is the project owner, so I suppose [this reply](https://github.com/lepoco/wpfui/pull/1030#issuecomment-2053958429) counts?

This PR is awesome though, thanks for creating it @jonlipsky.

pomianowski (approved):

PR #1067, opened 2024-04-22
Fixed the markup extension for the SymbolIcon and FontIcon class not working at all.

m0lDaViA:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently you cannot use a markup extension for symbolicons or fonticons because the constructor does not accept the arguments.

grafik

## What is the new behavior?

The markup extension for the symbolicon and fonticon class are now working with some small issue.
grafik

## Other information

So, I have now tested this extensively. I had previously closed this thread because for some reason the FontIcon markup extension randomly worked. However, I think that was due to VS (bugs and such).
I also confused the FontIcon class with the SymbolIcon class in the thread (sorry x.x).

So: The markup extension needs a standard constructor _without_ arguments. Furthermore, the existing constructor does not accept 2 arguments.
Because of the bugs in VS, I wasn't sure before whether it was really a mistake in the markup extension
(that's why i closed the last pull request), but now I am. The SymbolIcon markup extension had the same problem. 
The font size doesn't (forgot work), but it's not clear to me why.
Maybe you know why @pomianowski ?

### DIFF
--- a/src/Wpf.Ui/Controls/Arc/Arc.cs
+++ b/src/Wpf.Ui/Controls/Arc/Arc.cs
@@ -3,27 +3,24 @@
 // Copyright (C) Leszek Pomianowski and WPF UI Contributors.
 // All Rights Reserved.
 
+using System.Windows.Controls;
+using System.Windows.Shapes;
 using Point = System.Windows.Point;
 using Size = System.Windows.Size;
+#pragma warning disable SA1124
 
-// ReSharper disable once CheckNamespace
 namespace Wpf.Ui.Controls;
 
-/// <summary>
-/// Control that draws a symmetrical arc with rounded edges.
-/// </summary>
-/// <example>
-/// <code lang="xml">
-/// &lt;ui:Arc
-///     EndAngle="359"
-///     StartAngle="0"
-///     Stroke="{ui:ThemeResource SystemAccentColorSecondaryBrush}"
-///     StrokeThickness="2"
-///     Visibility="Visible" /&gt;
-/// </code>
-/// </example>
-public class Arc : System.Windows.Shapes.Shape
+public class Arc : Shape
 {
+    #region Declarations
+
+    private Viewbox? _rootLayout;
+
+    #endregion
+
+    #region Static Properties
+
     /// <summary>Identifies the <see cref="StartAngle"/> dependency property.</summary>
     public static readonly DependencyProperty StartAngleProperty = DependencyProperty.Register(
         nameof(StartAngle),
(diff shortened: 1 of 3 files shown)


pomianowski (approved):

m0lDaViA:
@pomianowski I have no idea why this commit went into this pull request :s
Sorry for that, but i coulnd't even create another one.

So: As the arc class wasn't quite working so i thought "Why not? I do need it anyways.".
The arc class had a few problems including that the arc had to be re-inserted (as a control) when the size was changed because a resize was not taken into account when rendering.
That was the biggest problem that made me want to adapt these in the first place.

In the course of this, I thought it would be better for the user if he had the opportunity to edit both the sweepdirection and the penlinecap.

I also changed the starting position, from 3 a.m. to 12 a.m. I'll be happy to explain the reason for this:
Ultimately, it is important to the user that it is as easy to use as possible. This also applies to programmers. And what's easier than a clock that you've known since birth?

If you don't agree and/or are satisfied with it, it can always be changed^^

src/Wpf.Ui/Controls/Arc/Arc.cs
```diff
@@ -3,27 +3,25 @@
...
-///     StartAngle="0"
-///     Stroke="{ui:ThemeResource SystemAccentColorSecondaryBrush}"
-///     StrokeThickness="2"
-///     Visibility="Visible" /&gt;
-/// </code>
-/// </example>
-public class Arc : System.Windows.Shapes.Shape
+public class Arc : Shape
 {
+    #region Declarations
```
pomianowski, line 17: Try to not use regions

src/Wpf.Ui/Controls/Arc/Arc.cs
```diff
@@ -3,27 +3,25 @@
...
+using System.Windows.Shapes;
 using Point = System.Windows.Point;
 using Size = System.Windows.Size;
+#pragma warning disable SA1124
+#pragma warning disable CS0108
 
-// ReSharper disable once CheckNamespace
 namespace Wpf.Ui.Controls;
 
-/// <summary>
```
pomianowski, line 12: Why removing docs?

pomianowski (changes requested):

src/Wpf.Ui/Controls/Arc/Arc.cs
m0lDaViA: Well i removed them for a better overview and just forgot to add them back x.x

src/Wpf.Ui/Controls/Arc/Arc.cs
m0lDaViA: I can remove them if you want, just tell me. I never understood why anyone shouldn't use them.

pomianowski: Purely stylistically, I think they are code smell. They hide part of the code because it is difficult to read, and if the code is difficult to read, it means that it is not good code. It is known that WPF UI needs a lot of work to improve readability and cleanliness, but it is worth working on it

pomianowski (approved):

ElectroAttacks:
@m0lDaViA  I found a new "feature" while using the NavigationView in LeftFluent mode. If you are using a different Font like Montserrat (which is actually quite big), you end up with a cut-off string, even if you increase the item size manually. I think it might be the ContentPresenter...

Issue #927, opened 2024-02-02
When I click the Switch Theme button Twice,the theme Mica is not work


Smileher:
### Describe the bug

When I click the Switch Theme button Twice,the theme Mica is not work

### To Reproduce

Switch Dark Theme

### Expected behavior

Mica

### Screenshots

![WeChat截图_20240202132206](https://github.com/lepoco/wpfui/assets/16812755/cbc6289a-94c0-4968-9e4f-6cd8c372d654)
![WeChat截图_20240202132505](https://github.com/lepoco/wpfui/assets/16812755/7420a9fd-b9c8-456d-bd66-569c65fd9622)
![WeChat截图_20240202133049](https://github.com/lepoco/wpfui/assets/16812755/fbd866a6-99e5-4b7a-a076-fa5683c2abe7)

### OS version

Windows 11  23H2

### .NET version

.NET FrameWork 4.8

### WPF-UI NuGet version

3.0.0-preview.13

johnvesalius:
Yeah, I see the same behaviour in 3.0.0. The Mica effect disappears when switching themes and is only applied again when the app is restarted.

Muniwedesu:
By the way, it doesn't work because there was a fix for another issue (ig), which disabled setting background to transparent. 
To get this to work, you need to set your window background to transparent (there's a public method for this, it only needs your window reference)
You can fix that like so:
```cs
{
    ApplicationThemeManager.Apply(appTheme == ApplicationTheme.Light
                    ? ApplicationTheme.Dark
                    : ApplicationTheme.Light);
                    // where 'this' is your window instance, 
                    // winver is client OS build number, but it doesn't seem to break anything 
                    // when you remove background on the earlier versions 
                    // 22621 is windows build number which has mica support, though it may also work fine on some earlier ones
                if (winver > 22621) WindowBackdrop.RemoveBackground(this); 
}
```

Smileher:
> 顺便说一句，它不起作用，因为修复了另一个问题（ig），该问题禁用了将背景设置为透明。要让它工作，你需要将你的窗口背景设置为透明（有一个公共方法，它只需要你的窗口引用）你可以像这样解决这个问题：
> 
> ```cs
> {
>     ApplicationThemeManager.Apply(appTheme == ApplicationTheme.Light
>                     ? ApplicationTheme.Dark
>                     : ApplicationTheme.Light);
>                     // where 'this' is your window instance, 
[...]

PR #1094, opened 2024-05-20
fix non-legacy backdrop when switching themes

Muniwedesu:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #927 

## What is the new behavior?

- backdrop is applied correctly

## Other information

checked on windows 11 22631 build and on windows 10 19045 build (only default light/dark themes with simple demo app)

### DIFF
--- a/src/Wpf.Ui/Appearance/WindowBackgroundManager.cs
+++ b/src/Wpf.Ui/Appearance/WindowBackgroundManager.cs
@@ -91,10 +91,14 @@ public static class WindowBackgroundManager
         }
 
         // This was required to update the background when moving from a HC theme to light/dark theme. However, this breaks theme proper light/dark theme changing on Windows 10.
-        // else
-        // {
-        //    _ = WindowBackdrop.RemoveBackground(window);
-        // }
+        // But window backdrop effects are not applied when it has an opaque (or any) background on W11 (so removing this breaks backdrop effects when switching themes), however, for legacy MICA it may not be required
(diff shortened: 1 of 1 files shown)


nabeelio:
I'm trying this out based on your last comment (nice timing!), but it's setting the entire window to transparent, instead of applying the Mica background to the navigation control

Muniwedesu:
@nabeelio afaik mica bg is drawn under the window (as this is set through winapi) and then it will be visible on different controls (including window) only when they either don't have background or the background is transparent, so window bg removal is intended.
Though, if it isn't displayed at all, you seem to have some other stuff happening (is your window becoming literally transparent or it just has black/white bg?). Firstly, you should probably check if your OS build supports this effect. It also seems to do nothing if you have backdrop effect set to "Auto", so make sure you set proper backdrop when applying theme.

nabeelio:
@Muniwedesu thanks - that's interesting. I'm on Windows 11. And yes, setting this, it applies the transparency to everything. I'm converting this over from ModernWPF (removed everything). My window backdrop type is set to Mica. I'm assuming it might be an `app.manifest` file problem

Muniwedesu:
@nabeelio maybe something else is changing background settings through winapi after you swap themes - wpfui sets your window background to black/white when you apply any theme, so the window should never become transparent.

nabeelio:
Yeah, maybe. It's very strange. I'll continue debugging, but good to know your fix works as intended! Thank you

Muniwedesu:
Check if you have these attributes set in your window tag 
image

pomianowski (approved):

PR #1092, opened 2024-05-18
Fixed menu item add when item added to the ObservableCollection

crippledfaith:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes

## What is the current behavior?

When MenuItemsSource or FooterMenuItemsSource in MVVM binding to ObservableCollection the Item added to the collection dose not update in the UI. 

Issue Number: N/A

## What is the new behavior?
```
        [ObservableProperty]
        private ObservableCollection<object> _broserNavigationViewItem = new ObservableCollection<object>();

         BroserNavigationViewItem.Add(new NavigationViewItem()
         {
             Content = content,
             Icon = new SymbolIcon { Symbol = icon },
             TargetPageType = targetPageType,
             TargetPageTag = name,
             Template = NavigationService.GetNavigationControl().ItemTemplate
         });
```
NavigationView item will add the Item

PR #1098, opened 2024-05-23
Update `INotifyCollectionChanged` events in `MenuItems` and `FooterMenuItems` of the `NavigationView`

ioswald:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

If you modify a databound observable collection to navigationview menu/footer items and then modify that list later in the program, the ui will not update with the new items.

Issue Number: N/A

## What is the new behavior?
Ui now updates when binding source is updated.

- added source binding list change event handling for navigationview menu/footer items.
- Added collection action handling of menu/footer items.
-
-

### DIFF
--- a/src/Wpf.Ui/Controls/NavigationView/NavigationView.Properties.cs
+++ b/src/Wpf.Ui/Controls/NavigationView/NavigationView.Properties.cs
@@ -468,8 +468,12 @@ public partial class NavigationView
 
             case NotifyCollectionChangedAction.Remove:
                 foreach (var item in e.OldItems)
+                {
                     if (!e.NewItems.Contains(item))
+                    {
                         collection.Remove(item);
+                    }
+                }
                 break;
 
             case NotifyCollectionChangedAction.Move:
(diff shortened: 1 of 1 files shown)


ioswald:
Fixed stackoverflow due to recursive call from collection change event when updating from binding source.

crippledfaith:
Hi please look at (https://github.com/lepoco/wpfui/pull/1092) You don't need to add the items to another collection as it is an observable collection, any change to the collection will change the items.

ioswald:
The initial items get added to the observable collection and update when you set the binding source, but you aren't adding a binding to the binding source so if the binding source changes after the initial binding is done; the items do not update.

src/Wpf.Ui/Controls/NavigationView/NavigationView.Properties.cs
```diff
@@ -450,6 +450,45 @@ public Thickness FrameMargin
...
+    {
+        if (ReferenceEquals(sender, collection))
+        {
+            return;
+        }
+
+        switch (e.Action)
+        {
+            case NotifyCollectionChangedAction.Add:
+                foreach (var item in e.NewItems)
```
pomianowski, line 463: Use explicit type

Issue #1043, opened 2024-04-09
PasswordBox Placeholder Not Working Properly


Taiizor:
### Describe the bug

PasswordBox Placeholder not working properly.

### To Reproduce

`<ui:PasswordBox
      Password="Sucrose Wallpaper Engine"
      PlaceholderEnabled="True"
      PlaceholderText="Enter a password" />`

### Expected behavior

When PlaceholderEnabled is true, Placeholder should not appear if there is data in the Password.

### Screenshots

![image](https://github.com/lepoco/wpfui/assets/41683699/db94da18-d1bb-402e-a969-d5257a744b77)

### OS version

Windows 11 24H2

### .NET version

.NET 9

### WPF-UI NuGet version

3.0.3

textGamex:
You can set `PlaceholderEnabled` to `false` to avoid this bug; this will not affect the normal display of `PlaceholderText`. I am looking for a solution to fix it.

Taiizor:
> You can set `PlaceholderEnabled` to `false` or remove `PlaceholderEnabled="True"` to avoid this bug; this will not affect the normal display of `PlaceholderText`. I am looking for a solution to fix it.

I'm already using this solution temporarily, but a permanent solution would be better (:

PR #1109, opened 2024-06-04
fix: TextBox, NumberBox, PasswordBox Placeholder Not Working Properly.

textGamex:
## Pull request type

- [ ] Update
- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [x] Documentation content changes

- When `Password` is set before setting `PlaceholderEnabled`, both `PlaceholderText` and `Password` will appear simultaneously.
- If `PlaceholderEnabled` is `false` and `PlaceholderText` is not empty, clearing Text will still display `PlaceholderText`.

Issue Number: #1043

## What is the new behavior?

- fix this bugs.
- If `PlaceholderEnabled` is set to `false` and `PlaceholderText` is not empty, clearing `Text` will not display `PlaceholderText`.

## Other information
Added a new dependency property `CurrentPlaceholderEnabledProperty`.


textGamex:
I don't understand why the default value of PlaceholderEnabled is true, and when we assign true to it again, both Password and TextBox will be displayed at the same time.

pomianowski (approved):

Issue #887, opened 2023-12-20
Button FontSize ineffective 


IOL0ol1:
### Describe the bug

<Button Content="Button1" FontSize="100" /> fontsize ineffective .
nuget vbersion Preview.9 is OK, Preview 10,11,12,13 all can be reproduced.

### To Reproduce

minimal  project ，then  <Button Content="Button1" FontSize="100" />

### Expected behavior

plz fixed it

### Screenshots

![image](https://github.com/lepoco/wpfui/assets/13666120/b6288f1c-dc83-4e71-9e77-6e1a5c10b937)

### OS version

windows 11

### .NET version

net6.0-windows

### WPF-UI NuGet version

3.0.0-preview.13
3.0.0-preview.12
3.0.0-preview.11
3.0.0-preview.10

### Additional context

Preview.9 is OK, Preview 10,11,12,13 all can be reproduced

NoThrottle:
Can confirm. `Font Size` affects icons instead. The "Font Size" property in `Icon="{ui:SymbolIcon Person24, FontSize=24}"` does not work as well.

IOL0ol1:
> Can confirm. `Font Size` affects icons instead. The "Font Size" property in `Icon="{ui:SymbolIcon Person24, FontSize=24}"` does not work as well.

Perhaps all Content has this issue

SouljaVR:
> > Can confirm. `Font Size` affects icons instead. The "Font Size" property in `Icon="{ui:SymbolIcon Person24, FontSize=24}"` does not work as well.
> 
> Perhaps all Content has this issue

seems so, can confirm it only works on icons and not button text

frostybee:
I am not experiencing any issue with changing the font size of any control. Perhaps you are not configuring the project the right way?

![image](https://github.com/lepoco/wpfui/assets/805978/bc7de48c-348c-4daa-85c7-31c73060a635)

IOL0ol1:
@frostybee here is my test VS solution: https://github.com/IOL0ol1/WpfUITest.git， 
What is the effect of your operation?

here is my MainWindow :
**Windows 11 Pro 22H2 (22621.3007)**
**WPF-UI 3.0.0-preview.13**

![image](https://github.com/lepoco/wpfui/assets/13666120/a05659fe-d812-4c0a-a4b2-ee29d2d54c60)

IOL0ol1:
@frostybee @SouljaVR @NoThrottle @timheuer hi, everyone. 

[...]

PR #1117, opened 2024-06-10
Added the HorizontalIconAlignment property to the button.

m0lDaViA:
Added the HorizontalIconAlignment property to the button.
You can now choose, if you want to show the icon left or right to the content.

## Pull request type

- [x] Update
- [x] Bugfix
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] Documentation content changes

Issue Number: N/A

## What is the new behavior?

-
-


src/Wpf.Ui/Controls/Button/Button.cs
```diff
@@ -13,16 +13,16 @@ namespace Wpf.Ui.Controls;
 /// </summary>
 /// <example>
 /// <code lang="xml">
-/// &lt;ui:Button
```
pomianowski, line 16: Examples should not use special characters
https://wpfui.lepo.co/api/Wpf.Ui.Controls.Button.html

src/Wpf.Ui/Controls/Button/Button.cs
```diff
@@ -181,6 +188,21 @@ public Brush PressedBorderBrush
...
+    /// <summary>
+    /// Gets or sets whether the icon should be shown left or right from the content.
+    /// </summary>
+    public HorizontalIconAlignments HorizontalIconAlignment
+    {
+        get => (HorizontalIconAlignments)GetValue(HorizontalIconAlignmentProperty);
+        set => SetValue(HorizontalIconAlignmentProperty, value);
+    }
+
+    public enum HorizontalIconAlignments
```
pomianowski, line 203: Missing docs for public API.

pomianowski, line 203: I would extract it to a separate file

pomianowski (changes requested):

src/Wpf.Ui/Controls/Button/Button.cs
m0lDaViA: Alright.

src/Wpf.Ui/Controls/Button/Button.cs
m0lDaViA: Because? I mean even vs is using it totally fine.

m0lDaViA: Oh and btw & is a special character :)

src/Wpf.Ui/Controls/Button/Button.cs
m0lDaViA: Any suggestion for the docs?

Jay-o-Way:
Hm, the weird "alignment" usually should like it is placed inside of a space. Like: text on a page is aligned inside/between the borders of the page. Not _next to_ the page.
Here, is more like a position _relative_ to something else.

m0lDaViA:
> Hm, the word "alignment" usually should like it is placed inside of a space. Like: text on a page is aligned inside/between the borders of the page. Not _next to_ the page. Here, is more like a position _relative_ to something else.

But it isn't relative, that's the difference.

Jay-o-Way:
What I mean is, the icon is placed to the left/right of the content/text. Therefore I think "HorizontalIconPlacement" would be a good (or even more accurate) wording.

By the way, I assume this works well in RTL language?

m0lDaViA:
> What I mean is, the icon is placed to the left/right of the content/text. Therefore I think "HorizontalIconPlacement" would be a good (or even more accurate) wording.
> 
> By the way, I assume this works well in RTL language?

In this case I followed the standard alignment that MS uses.

m0lDaViA:
#1120 & #887 can check the last commit.

m0lDaViA:
I don't know why, but my vs is bugged rn. I cannot create visible styles and my formatter does instead of formatting only opened files all .xaml files.
Sorry for that.

PR #1122, opened 2024-06-10
Fix titlebar color when using backdrop in WPF `Window`

pomianowski:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Window can have a color in custom titlebar when **Show accent colors** option is enabled in Windows.

Issue Number: N/A

## What is the new behavior?

After enabling Mica backdrop, the titlebar color should be removed.

### DIFF
--- a/src/Wpf.Ui/Appearance/WindowBackgroundManager.cs
+++ b/src/Wpf.Ui/Appearance/WindowBackgroundManager.cs
@@ -100,6 +100,7 @@ public static class WindowBackgroundManager
         }
 
         _ = WindowBackdrop.ApplyBackdrop(window, backdrop);
+
         if (applicationTheme is ApplicationTheme.Dark)
         {
             ApplyDarkThemeToWindow(window);
@@ -108,6 +109,8 @@ public static class WindowBackgroundManager
         {
             RemoveDarkThemeFromWindow(window);
         }
+
+        _ = WindowBackdrop.RemoveTitlebarBackground(window);
 
         foreach (object? subWindow in window.OwnedWindows)
         {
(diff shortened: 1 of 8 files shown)


pomianowski:
FYI: @Jay-o-Way, @aquinn39
rel.: https://github.com/microsoft/PowerToys/issues/30206

jaimecbernardo:
@pomianowski  Awesome work here :)
I've added this fix to https://github.com/microsoft/PowerToys/pull/33458 , while 4.0.0 is not released.
Thank you!

Quick question: Why `uint titlebarPvAttribute = 0xFFFFFFFE;` and not `uint titlebarPvAttribute = 0xFFFFFFFF;` ?

aquinn39:
> @pomianowski  Awesome work here :)
> 
> I've added this fix to https://github.com/microsoft/PowerToys/pull/33458 , while 4.0.0 is not released.
> 
> Thank you!
> 
> 
> 
> Quick question: Why `uint titlebarPvAttribute = 0xFFFFFFFE;` and not `uint titlebarPvAttribute = 0xFFFFFFFF;` ? 

I got the value from DWMWA_BORDER_COLOR [here](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute). This value appearantly hides the window border, so I thought maybe it might work for the title bar as well, which it did. 0xFFFFFFFF (DWMWA_COLOR_DEFAULT) just resets the title bar colour to the default system behaviour, which means using the accent colour if you have that option enabled - that behaviour actually is documented under DWMWA_CAPTION_COLOR.

Makes me wonder if DWMWA_COLOR_NONE (0xFFFFFFFE) also works on DWMWA_TEXT_COLOR to hide the caption text (not that it's necessary in this case).

aquinn39:
Interestingly this behaviour is actually correctly documented in dwmapi.h.
image

Issue #447, opened 2022-10-27
Card control hover and click


Pietro228:
### Is your feature request related to a problem? Please describe

Card control has hover and press effect but if it's not used for clicking it should do these effects. This could confuse users.

### Describe the solution you'd like

Make an option to remove these hover and press effects.

### Describe alternatives you've considered

Use card with grid that has columns etc. but it takes too much space in XAML and it's time consuming.

PR #1233, opened 2024-09-24
Fix hover behavior for CardControl

Difegue:
Non-actionable cards shouldn't have hover behavior, as per WinUI

## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

CardControls behave as if they can be clicked when hovering over them

Issue Number: #447

## What is the new behavior?

- CardControls are inert, as in the Windows Settings app

### DIFF
--- a/src/Wpf.Ui/Controls/CardControl/CardControl.xaml
+++ b/src/Wpf.Ui/Controls/CardControl/CardControl.xaml
@@ -86,24 +86,6 @@
                         </Grid>
                     </Border>
                     <ControlTemplate.Triggers>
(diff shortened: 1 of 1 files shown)


pomianowski (approved):

seikosantana:
In the WPF-UI Gallery it is still clickable

seikosantana:
Hi, might be not the correct way to use the CardControl, but I like how it has Icon and header slot out of the box and I'd like to use it like a button.

Is there any recommended way to enable the hover and click behavior for CardControl?

Thanks in advance

PR #1244, opened 2024-10-01
Defining a FontSize in a FontIcon should not be overridden by its VisualParent's

Difegue:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using a `FontIcon`, its `FontSize` will first take the FontSize of its `VisualParent` if it has one, then be set to `DefaultIconFontSize` if it's still at the default WPF FontSize.   

Both of those actions override whatever FontSize you would set in the FontIcon control itself.  
An easy example of this is if I want to have a `HyperlinkButton` with a custom Icon Font Size:  

```
<ui:HyperlinkButton x:Name="UpgradeHyperlink"
       Margin="0,2,0,0" Height="30"
       Icon="{ui:SymbolIcon ChevronCircleUp20, FontSize=20}"
       Content="Upgrade"
       />
```

image

Issue Number: N/A

## What is the new behavior?

The VisualParent FontSize override now only happens if the FontIcon was using the default WPF FontSize. (as in, no custom sizes were applied)  

image  

## Other information

There's arguably still a problem with this fix if you want a FontIcon whose size coincidentally matches `SystemFonts.MessageFontSize` within say, a HyperLinkButton with a FontSize of 20. But that'd require deeper architectural changes to handle properly...

### DIFF
--- a/src/Wpf.Ui/Controls/IconElement/FontIcon.cs
+++ b/src/Wpf.Ui/Controls/IconElement/FontIcon.cs
@@ -117,7 +117,7 @@ public class FontIcon : IconElement
             SetResourceReference(FontSizeProperty, "DefaultIconFontSize");
 
             // If the FontSize is the default, set it to the parent's FontSize.
-            if (VisualParent is not null)
+            if (VisualParent is not null && TextElement.GetFontSize(VisualParent) != SystemFonts.MessageFontSize)
             {
                 SetCurrentValue(FontSizeProperty, TextElement.GetFontSize(VisualParent));
             }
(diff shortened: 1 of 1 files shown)


Namax-uniman:
is this fixed? or are there any way to increase icon size without increasing font size. i am using icon in textbox and the only way to increase the size is by increasing text font size.

Difegue:
with this PR, you can fully decorrelate icon fontsize from other text.

pomianowski (approved):

PR #1344, opened 2025-02-10
Upgraded VS Extension to use WPF-UI 4.0.0

Djohnnie:
## Pull request type

- [x] Update
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The current VS extension uses the deprecated [3.0.1](https://www.nuget.org/packages/WPF-UI/3.0.1) version of WPF-UI.

Issue Number: N/A

## What is the new behavior?

This updated VS extensions uses the new [4.0.0](https://www.nuget.org/packages/WPF-UI/4.0.0) version of WPF-UI.

## Other information

I hope to have fixed the breaking changes between version 3.0.1 and version 4.0.0. I've tested the three project templates and they look to be working as expected.


src/Wpf.Ui.Extension.Template.Blank/App.xaml.cs
```diff
@@ -27,22 +27,19 @@ public partial class App
...
         }
 
         /// <summary>
         /// Occurs when the application is loading.
         /// </summary>
-        private void OnStartup(object sender, StartupEventArgs e)
+        private async void OnStartup(object sender, StartupEventArgs e)
         {
-            _host.Start();
+            await _host.StartAsync();
```
chucker, line 42: Consider wrapping this in `try/catch` due to `async void`.

Djohnnie: Any suggestions on what to do if an Exception got caught? I basically copied this approach from the Wpf.Ui.Demo.Mvvm project and I'm not sure what else we can do when application initialization and startup fail.
We could call Environment.Exit(1)?

chucker: You're right, there's not _much_ that can be done.

I _believe_ not catching it will in this case just send the exception to Windows' Event Log, so the user may be confused why the app is just gone. IMHO, you should present the exception in a plain `MessageBox`, so it's at least clear that a crash occurred. (WinForms has a built in exception dialog, but I believe WPF does not.)

pomianowski (approved):

PR #1373, opened 2025-03-04
TitleBar Buttons not working in some cases

Alessio2405:
This fix addresses an issue that occurred when detecting whether the mouse was hovering over one of the TitleBar buttons (such as minimize, maximize, or close) of a window. 
The previous method of detecting the mouse position could lead to incorrect results, particularly when the pointer was moved quickly from outside to inside the window.

## Pull request type

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

When creating an instance of a window with a valid parent and displaying it using ShowDialog (I can reproduce this issue only with modal windows), there's sometimes a behavior where you are unable to click any of the titlebar buttons.
 The only way to resolve this issue is by first giving focus to the content of the window and then returning focus to the titlebar area, making it difficult to interact with the titlebar buttons.

Issue Number: N/A

## What is the new behavior?

- Fixed MouseOver detection: The logic for detecting mouse hover over the TitleBar buttons has been improved. The method now consistently detects the mouse position by using an interop function to ensure the accuracy of this calculation.
- Correct transformation of coordinates: The transformation from screen coordinates to local window coordinates is now checked, even when device scaling or window transformations are in place, ensuring accurate hit testing on the TitleBar buttons.

## Other information

N/A

### DIFF
--- a/src/Wpf.Ui/Controls/TitleBar/TitleBarButton.cs
+++ b/src/Wpf.Ui/Controls/TitleBar/TitleBarButton.cs
@@ -232,4 +232,39 @@ public class TitleBarButton : Wpf.Ui.Controls.Button
             ),
         };
     }
+
+    PresentationSource presentationSource = null;
+
+	protected bool IsMouseOverElement(nint lParam)
+	{
+		System.Drawing.Point winPoint;
+		bool gotCursorPos = User32.GetCursorPos(out winPoint);
+
+		if (!gotCursorPos)
+		{
+			int fallbackX = unchecked((short)((long)lParam & 0xFFFF));
+			int fallbackY = unchecked((short)(((long)lParam >> 16) & 0xFFFF));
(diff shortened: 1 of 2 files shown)


Difegue:
This doesn't work in the gallery on my end - The screen coordinates seem to incorrectly translate to the visual's.

Alessio2405:
> This doesn't work in the gallery on my end - The screen coordinates seem to incorrectly translate to the visual's.

High-dpi issue or something? If you try something like 

`var screenPointLogical = presentationSource.CompositionTarget.TransformFromDevice.Transform(
		new System.Windows.Point(winPoint.X, winPoint.Y));

var localPoint = this.PointFromScreen(screenPointLogical);`

before translating point from screen coordinates does it work for you?

I'm gonna post a gif with the behaviour before and after, since it seemed to be ok in my tests

pomianowski (approved):

Issue #1434, opened 2025-05-26
Update `System.ValueTuple` dependency from `4.5.0` to `4.6.1`.


mjr4077au:
### Describe the bug

A member of our community [reported](https://discourse.psappdeploytoolkit.com/t/defender-for-cloud-apps-alerts-on-system-valuetuple-dll/6552) that their endpoint protection software flagged `System.ValueTuple.dll` as malicious, despite it being signed off by Microsoft, etc.

Given Microsoft has [recently put out `4.6.1`](https://www.nuget.org/packages/System.ValueTuple/4.6.1), maybe there was some issue with the file after all these years.

I've updated subtree clone of Wpf.Ui to use `4.6.1` [here](https://github.com/PSAppDeployToolkit/PSAppDeployToolkit/commit/1fcfa535277243444d48074fe9dbd13a5b80e44c) and have encountered no ill effects or issues following this change.

### To Reproduce

1. Use Microsoft Defender for Cloud Apps
2. Witness the apprent issue

### Expected behavior

1. Use any endpoint protection software
2. Experience no issues with MSFT-signed dependencies

### OS version

Windows

### .NET version

net462

### WPF-UI NuGet version

4.0.0

jbryknar:
I recommend reporting this incident to Microsoft as a false positive. I don't think, even after a scan with corporate/cloud antiviruses, that this particular library contains a bad code sequence that would indicate that it is, for example, a patch, crack or other trojan.

https://learn.microsoft.com/en-us/defender-xdr/m365d-autoir-report-false-positives-negatives#report-a-false-positivenegative-to-microsoft-for-analysis

I think we can upgrade the library/ NuGet package System.ValueTuple to the latest version, it's not against anything. I'll create a PR for it and we can throw it into the main branch. 

Edit: PR #1450

PR #1450, opened 2025-06-13
Update System.ValueTuple dependency from 4.5.0 to 4.6.1

jbryknar:
## Pull request type

- [X] Update
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

A user @mjr4077au reported that their endpoint protection software incorrectly flagged the System.ValueTuple.dll—a Microsoft-signed library—as malicious. This occurred despite the library being widely trusted and in use for years. The issue was raised in the context of updating the System.ValueTuple package to version 4.6.1. 

Issue Number: #1434

## What is the new behavior?

The System.ValueTuple package has been updated to version 4.6.1 within the Wpf.Ui subtree. This change helps ensure compatibility with the latest security standards and mitigates false-positive detections by endpoint protection software. Following the update, the library continues to function as expected without introducing regressions or runtime issues.

image

## Other information
N/A


pomianowski:
Hey @jbryknar,
Thanks for your contributions to WPF UI and your PRs. Updating dependencies is generally a good practice, but in my view, we should aim to stick to the lowest reasonable version of dependencies whenever possible. In large organizations, solutions often consist of dozens of projects and thousands of dependencies. If even one project has a strict requirement on an older version of ValueTuple, upgrading can become difficult or even unfeasible.

For others still using .NET Framework 4.6.2 (which is common in many legacy systems), if they want to update, it's usually sufficient for the package consumer to manually reference a higher version of System.ValueTuple. See:

https://learn.microsoft.com/en-us/nuget/concepts/dependency-resolution#lowest-applicable-version

https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort

Always happy to hear your thoughts.
Regards,
Leszek

jbryknar:
Hi @pomianowski 

Thank you for your thoughtful feedback — I completely understand your point, especially when it comes to maintaining compatibility in larger solutions that still rely on older frameworks.

That said, in the context of WPF UI, I personally believe the focus should now shift more towards the modern .NET platform. We should aim to stick to long-term supported versions and keep things moving forward. At some point, these changes are inevitable — and while I agree that this specific case may not introduce a vulnerability, sticking with the lowest possible versions indefinitely can expose us to other risks down the line.

I absolutely respect your perspective and appreciate the discussion. Thanks.
Jan Bryknar

chucker:
>I personally believe the focus should now shift more towards the modern .NET platform.

Personally, I mostly use it from .NET Framework 4.7.2, which is still supported with `System.ValueTable` `4.6.1`, so I wouldn't be affected.

However, as far as the _modern_ .NET platform goes, one might ask: why not use WinUI 3 itself, then? I would wager a significant portion of users of WPFUI are .NET Framework users for this reason.

Issue #1250, opened 2024-10-10
Paste issue with PasswordBox and asterisk *


timokoessler:
### Describe the bug

If you paste text containing the asterisk (`*`) character inside a PasswordBox and click on the eye icon to view the text, the asterisk is removed. This only seems to happen when the password is pasted.

### To Reproduce

Paste `test*` to a PasswordBox with hidden input. Click on the eye icon to show the input. The asterisk is removed from the element.

### Expected behavior

The input should not be modified.

### OS version

Windows 11 23H2 22631.4169

### .NET version

8.0

### WPF-UI NuGet version

3.0.5

### Additional context

https://github.com/timokoessler/2FAGuard/issues/43

jbryknar:
The PR #1455 was created based on this issue.  Thanks @timokoessler.

timokoessler:
Thanks a lot @jbryknar for creating a PR!

Issue #1148, opened 2024-07-01
Unexpected icon behavior with multiple PasswordBoxes


mizydorczyk:
### Describe the bug

With multiple PasswordBox controls, changing the state of one causes the state of the others to change. As in the video below.

### To Reproduce

```
<StackPanel>
    <ui:PasswordBox Margin="5" />
    <ui:PasswordBox Margin="5" />
    <ui:PasswordBox Margin="5" />
</StackPanel>
```

### Expected behavior

Icon should not disappear

### Screenshots

https://github.com/lepoco/wpfui/assets/74381129/0772fb63-d496-4ac2-a7be-bfd4b612b78b

### OS version

Windows 11 23H2

### .NET version

NET 8.0

### WPF-UI NuGet version

3.0.4

jbryknar:
The PR #1455 was created based on this issue.

PR #1455, opened 2025-06-16
Fix paste issue with PasswordBox and asterisk

jbryknar:
## Pull request type

- [X] Update
- [X] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

- When pasting or typing the `PasswordChar` (e.g. " * ") into the `PasswordBox`, the character is incorrectly discarded.
- This occurs because the input processing logic assumes any occurrence of the `PasswordChar` in the Text property is a masking symbol and not actual user input
-  As a result, characters matching the masking symbol cannot be used in passwords, and any pasted text containing such characters gets silently altered. Revealing the password also shows the text without the originally entered or pasted *, leading to data loss and unexpected behavior.

#### Before Fix
PasswordBox_BeforeFix

Issue Number: #1250 and #1148

## What is the new behavior?

The `PasswordBox` now properly distinguishes between masking and actual input. Characters identical to the `PasswordChar`, such as " * ", can now be entered or pasted as part of the password. 

The internal logic no longer strips or ignores these characters during input or paste operations. When the password is revealed, it correctly displays all characters, including those matching the mask symbol.

#### After Fix

PasswordBox_AfterFix

#### Fixed unexpected icon behavior with multiple PasswordBoxes

PasswordBox_AfterFix_Issue_1148

## Other information

N/A

#### Testing Scenarios
Scenario | Scenario Steps | Expected Result | Status
--- | --- | --- | ---
Type text with special characters in revealed mode | In PasswordBox type `Test#*{!ˇ12'` | The entered password will be displayed in the field | Passed
Type text with special characters in hidden mode | In PasswordBox type `Test#*{!ˇ12'` | Wildcards will be displayed in the field. When you click on the reveal password button, the password entered by the user should be displayed. | Passed
Testing Copy & Paste Action | Write the password somewhere on the side (for example, Notepad) and then copy and paste it into the box. The password must contain special characters, including the wildcard character “ * ”.  For example: `Test*12@#!!(` Paste this password into text field. | After pasting and revealing the password with the push button, the password must be complete as it was originally copied to the clipboard. | Passed
Testing Copy & Paste with Replace All Action | Write the password somewhere on the side (for example, Notepad) and then copy and paste it into the box. The password must contain special characters, including the wildcard character “ * ”.  For example: `Test*12@#!!(` Paste this password into field. Select all text (CTRL+A or manually with the cursor) and re-enter the password from clipbaord |  After pasting, clearing and revealing the password with the push button, the password must be complete as it was originally copied to the clipboard. | Passed

### DIFF
--- a/src/Wpf.Ui/Controls/PasswordBox/PasswordBox.cs
+++ b/src/Wpf.Ui/Controls/PasswordBox/PasswordBox.cs
@@ -93,7 +93,7 @@ public partial class PasswordBox : TextBox
     }
 
     /// <summary>
-    /// Gets a value indicating whether gets whether the password is currently revealed.
+    /// Gets a value indicating whether the password is currently revealed.
     /// </summary>
     public bool IsPasswordRevealed
     {
@@ -102,7 +102,7 @@ public partial class PasswordBox : TextBox
     }
 
     /// <summary>
-    /// Gets or sets a value indicating whether gets or sets whether the password reveal button is enabled.
+    /// Gets or sets whether the password reveal button is enabled.
     /// </summary>
     public bool RevealButtonEnabled
     {


src/Wpf.Ui/Controls/PasswordBox/PasswordBox.cs
```diff
@@ -80,7 +93,7 @@ public char PasswordChar
     }
 
     /// <summary>
-    /// Gets a value indicating whether the password is revealed.
+    /// Gets a value indicating whether gets whether the password is currently revealed.
```
pomianowski, line 96:
```suggestion
    /// Gets a value indicating whether the password is currently revealed.
```

pomianowski (approved):

Nuklon:
This logic is completely broken, I'll create a new PR to address it.

Nuklon:
https://github.com/lepoco/wpfui/pull/1547

PR #1366, opened 2025-02-27
TitleBar Header and TrailingContent IsHitTestVisible

xylieong:
…e, then user is not able to click on the UIElement.

When TitleBar is customised with TextBlock only and with IsHitVisible set false, then allow user to click on the TextBlock and able to select & move the window.

## Pull request type

Please check the type of change your PR introduces:

- [/ ] Update
- [ ] Bugfix
- [/ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes

## What is the current behavior?

TitleBar Header and TrailingContent makes user unable to click to select window and move window.

Issue Number: N/A

## What is the new behavior?

If set 'IsHitTestVisible' of UIElement of Header or TrailingContent to False, then user will not be able to hit on the UIElement and be able to select the window (and move the window, if they want to).

## Other information

N/A

Issue #1126, opened 2024-06-11
Window caption buttons don't work properly with touch


Difegue:
### Describe the bug

When using touch to interact with a WPFUI app, tapping the maximize button seems to fire two actions, which maximizes then restores the window to its previous size.  

This worked fine in 2.x.

### To Reproduce

Open the controls app on a touchscreen-equipped PC and tap the maximize button

### Expected behavior

The app should maximize properly

### Screenshots

https://github.com/lepoco/wpfui/assets/8237712/80edefee-7812-4c92-8981-38188f24cc1a

### OS version

Win11

### .NET version

Any version

### WPF-UI NuGet version

3.0.3

PR #1460, opened 2025-06-24
Bring back the workaround for #1126

Difegue:
This brings back the workaround for touch input triggering the maximize button twice that got removed by #1366 .

## Pull request type

- [ ] Update
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Using a touchscreen to interact with the maximize button in an app using WPFUI will trigger it twice, maximizing and restoring the window in quick succession. 

Issue Number: #1126

## What is the new behavior?

Touch only maximizes or restores the app once.

### DIFF
--- a/src/Wpf.Ui/Controls/TitleBar/TitleBar.xaml
+++ b/src/Wpf.Ui/Controls/TitleBar/TitleBar.xaml
@@ -173,11 +173,13 @@
                                     Grid.Column="1"
                                     ButtonType="Minimize" />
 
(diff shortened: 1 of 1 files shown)


Difegue:
Nevermind, seems like this doesn't actually work anymore due to changes to the hwndhook. Will have to dig deeper...

ctmusall:
> Nevermind, seems like this doesn't actually work anymore due to changes to the hwndhook. Will have to dig deeper...

Any updates on fixes for this issue?

Difegue:
nope! help welcome :)

PR #1473, opened 2025-07-04
Fix button brushes with appreance

Nuklon:
## Pull request type

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Fixes 2 issues.
1. Currently the Primary appreance button sort of gets moved up due to missing MouseOverBrush. You can see this here:
image
Right one has mouse over.
2. Adds an alternative border thickness (0,0,0,1) for buttons with colors. Right now it used 1,1,1,1 but this 
a. doesn't match WinUI style
b. causes the border shadow to flow over the sides. Fixed version:
image

### DIFF
--- a/src/Wpf.Ui/Controls/Button/Button.xaml
+++ b/src/Wpf.Ui/Controls/Button/Button.xaml
@@ -15,6 +15,7 @@
 
     <Thickness x:Key="ButtonPadding">11,5,11,6</Thickness>
     <Thickness x:Key="ButtonBorderThemeThickness">1</Thickness>
+    <Thickness x:Key="ButtonBorderThemeAltThickness">0,0,0,1</Thickness>
     <Thickness x:Key="ButtonIconMargin">0,0,8,0</Thickness>
 
     <Style x:Key="DefaultRepeatButtonStyle" TargetType="{x:Type RepeatButton}">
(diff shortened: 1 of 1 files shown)


Nuklon:
3. Also fixed the border brush when selected. With WinUI, the pressed border brush is identical to the background of the button, (and not transparent).

Nuklon:
Fixed remaining colors, they now look almost identical to WinUI. Depends on #1492.

Nuklon:
Comparison (WinUI left, WPFUI right):
image

src/Wpf.Ui/Controls/Button/Button.xaml
```diff
@@ -232,6 +233,7 @@
         <Setter Property="CornerRadius" Value="{DynamicResource ControlCornerRadius}" />
         <Setter Property="Appearance" Value="Secondary" />
         <Setter Property="SnapsToDevicePixels" Value="True" />
+        <Setter Property="UseLayoutRounding" Value="True" />
```
Nuklon, line 236: This is necessary to avoid the border thickness going into gray territory when you're not using 100% display scaling.

Nuklon:
New version added with an inset border, which is sort of what WinUI also does.

Nuklon:
https://github.com/lepoco/wpfui/pull/1473/commits/7a3eff8079e3adf3100be5c8a2e8e76755076380 fixes dark mode secondary buttons having persistent border. Also added some comments to clarify.

Issue #1481, opened 2025-07-10
ApplicationAccentColorManager has different logic than Accent.xaml


labsin:
### Describe the bug

In ApplicationAccentColorManager, the colors are set using this logic:
```
        UiApplication.Current.Resources["SystemAccentColor"] = systemAccent;
        UiApplication.Current.Resources["SystemAccentColorPrimary"] = primaryAccent;
        UiApplication.Current.Resources["SystemAccentColorSecondary"] = secondaryAccent;
        UiApplication.Current.Resources["SystemAccentColorTertiary"] = tertiaryAccent;

        UiApplication.Current.Resources["SystemAccentBrush"] = secondaryAccent.ToBrush();
        UiApplication.Current.Resources["SystemFillColorAttentionBrush"] = secondaryAccent.ToBrush();
        UiApplication.Current.Resources["AccentTextFillColorPrimaryBrush"] = tertiaryAccent.ToBrush();
        UiApplication.Current.Resources["AccentTextFillColorSecondaryBrush"] = tertiaryAccent.ToBrush();
        UiApplication.Current.Resources["AccentTextFillColorTertiaryBrush"] = secondaryAccent.ToBrush();
        UiApplication.Current.Resources["AccentFillColorSelectedTextBackgroundBrush"] =
            systemAccent.ToBrush();
        UiApplication.Current.Resources["AccentFillColorDefaultBrush"] = secondaryAccent.ToBrush();

        UiApplication.Current.Resources["AccentFillColorSecondaryBrush"] = secondaryAccent.ToBrush(0.9);
        UiApplication.Current.Resources["AccentFillColorTertiaryBrush"] = secondaryAccent.ToBrush(0.8);
```

Now looking at Accent.xaml, the defaults are set using:

```
    <!--  Colors depending on the theme should be changed by the Manager  -->
    <Color x:Key="SystemAccentColor">#3379d9</Color>

    <!--  While the name remains Light to stay with the official nomenclature, it's made dark in Dark Theme  -->

    <!--  SystemAccentColorDark1 | SystemAccentColorLight1  -->
    <Color x:Key="SystemAccentColorPrimary">#559ce4</Color>
    <!--  SystemAccentColorDark2 | SystemAccentColorLight2  -->
    <Color x:Key="SystemAccentColorSecondary">#80b9ee</Color>
    <!--  SystemAccentColorDark3 | SystemAccentColorLight3  -->
[...]

PR #1492, opened 2025-07-17
fix(controls): Wrong accent colors

Nuklon:
## Pull request type

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

The current brushes are all wrong and don't match with WinUI:
image

Fixes: #1481
## What is the new behavior?

Gets the correct brushes from Windows. I also improved the fallback colors (still not perfect, but better).

image

WinUI gallery original colors (now WPF-UI has the same colors):
image

### DIFF
--- a/src/Wpf.Ui/Controls/Button/Button.xaml
+++ b/src/Wpf.Ui/Controls/Button/Button.xaml
@@ -245,34 +245,43 @@
                         HorizontalAlignment="{TemplateBinding HorizontalAlignment}"
                         VerticalAlignment="{TemplateBinding VerticalAlignment}"
                         Background="{TemplateBinding Background}"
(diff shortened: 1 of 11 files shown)


Nuklon:
Last 3 commits use the official documented accent color function, and remove the undocumented DwmGetColorizationParameters, which is unreliable on one of my machines (The system cannot find the file specified. (Exception from HRESULT: 0x80070002)). The register fallback works there. 
I've also now used AccentColor in the registery as this is not a reversed color.

chucker:
I haven't dug too deep, but this may be the cause of #1580. I'm guessing some of the changes to `ApplicationAccentColorManager` now take precedence over application-specified accent colors.

PR #1501, opened 2025-08-05
Fix PasswordBox not updating placeholder and clear button when initialized

Nuklon:
## Pull request type

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

If you use a PasswordBox, and specify a default password (Password="Test" for example), or bind it to an existing password, the placeholder doesn't get removed and the clear button isn't updated.

image

## What is the new behavior?

image

As one would expect.


src/Wpf.Ui/Controls/PasswordBox/PasswordBox.cs
```diff
@@ -123,12 +123,12 @@ public event RoutedEventHandler PasswordChanged
     protected override void OnTextChanged(TextChangedEventArgs e)
     {
         UpdateTextContents(isTriggeredByTextInput: true);
+        SetPlaceholderTextVisibility();
```
Nuklon, line 126: This should always be called to ensure the placeholder/clear button is in sync with the text.

Nuklon:
2nd commit centers the buttons vertically, same as TextBox, so they look largely identical again.

pomianowski (approved):

Nuklon:
Conflicts resolved.

PR #1505, opened 2025-08-07
Fix FluentWindow background with None backdrop extending into window

Nuklon:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes

## What is the current behavior?

Due to using `GlassFrameThickness` of `-1`, it looks like the background extends into the Window with `WindowBackdropType.None`.

image

## What is the new behavior?
Use a `GlassFrameThickness` of `1` if `WindowBackdropType.None` is used.

PR #1508, opened 2025-08-08
feat(controls): Add setting border color + fix background extending inside `FluentWindow`

Nuklon:
## Pull request type

- [ ] Update
- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently, WPF-UI doesn't set the border color based on accent. This adds that.
Old left, right new.

image

### DIFF
--- a/src/Wpf.Ui/Interop/UnsafeNativeMethods.cs
+++ b/src/Wpf.Ui/Interop/UnsafeNativeMethods.cs
@@ -85,8 +85,7 @@ public static class UnsafeNativeMethods
     /// <returns><see langword="true"/> if invocation of native Windows function succeeds.</returns>
     public static bool ApplyBorderColor(IntPtr handle, Color color)
     {
-        int colorNum = (color.B << 16) | (color.G << 8) | color.R;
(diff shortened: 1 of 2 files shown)


src/Wpf.Ui/Interop/UnsafeNativeMethods.cs
```diff
@@ -93,10 +122,7 @@ public static bool RemoveWindowDarkMode(IntPtr handle)
             dwAttribute = Dwmapi.DWMWINDOWATTRIBUTE.DMWA_USE_IMMERSIVE_DARK_MODE_OLD;
         }
 
-        // TODO: Validate HRESULT
```
Nuklon, line 96: I also corrected this, it's only returning 0 on success.

Nuklon:
Border is now only active when focused (same as WinUI). When deactivated it gets back to default OS color.

Nuklon:
This also fixes FluentWindow background with None backdrop extending into window. Taken from https://github.com/lepoco/wpfui/pull/1505.

pomianowski (approved):

PR #1511, opened 2025-08-11
Add CheckBox check animation

Nuklon:
## Pull request type

- [ ] Update
- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently the animation used by WinUI when a checkbox is being checked is missing.
This adds that

https://github.com/user-attachments/assets/8f808e39-8ff7-433d-9a89-555697256060

(Big example checkbox so you can see what's happening).

There's no animation when going from checked to indeterminate/checked to unchecked/indeterminate to unchecked, this is the same with WinUI.


Nuklon:
Going to close this for now. This should only be animated when the state changes, not when the initial state is checked.

PR #1512, opened 2025-08-13
feat(controls): Add `CheckBox` animation 

Nuklon:
## Pull request type

- [ ] Update
- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Replacement of https://github.com/lepoco/wpfui/pull/1511. The animation I've added is a little different from WinUI, which slides in from left to right. This is quite difficult to recreate in WPF. My initial idea to animate based on CheckBoxIconSize doesn't work reliably given the actual size is calculated based on the underlying glyph / TextBlock, and is sometimes larger. The new animation is now just a fade in based on opacity.

https://github.com/user-attachments/assets/8f808e39-8ff7-433d-9a89-555697256060

### DIFF
--- a/src/Wpf.Ui/Controls/CheckBox/CheckBox.xaml
+++ b/src/Wpf.Ui/Controls/CheckBox/CheckBox.xaml
@@ -134,7 +134,7 @@
                             <Setter TargetName="ControlBorderIconPresenter" Property="Background" Value="{DynamicResource CheckBoxCheckBackgroundFillChecked}" />
                             <Setter TargetName="StrokeBorder" Property="BorderBrush" Value="{DynamicResource CheckBoxCheckBorderBrush}" />
                             <Trigger.EnterActions>
-                                <BeginStoryboard Name="FadeIn">
+                                <BeginStoryboard Name="SlideIn">
                                     <Storyboard>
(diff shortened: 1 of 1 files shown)


Nuklon:
Thinking about it a little more, we can still do the idea of https://github.com/lepoco/wpfui/pull/1511, but now with the right properties to take the logical width into account.

jschroedl:
@Nuklon I see that the `CheckBoxAnimationFactorToValueConverter` added (line 37) is not being used by the XAML. Is that a remnant or am I missing something?

`  <converters:AnimationFactorToValueConverter x:Key="CheckBoxAnimationFactorToValueConverter" />`

Thanks a bunch for adding the animation BTW.

Nuklon:
> @Nuklon I see that the `CheckBoxAnimationFactorToValueConverter` added (line 37) is not being used by the XAML. Is that a remnant or am I missing something?
> 
> ` <converters:AnimationFactorToValueConverter x:Key="CheckBoxAnimationFactorToValueConverter" />`
> 
> Thanks a bunch for adding the animation BTW.

Yes, thanks, this is indeed unused, I've created a PR to drop this.

PR #1472, opened 2025-07-03
fix(controls): Tooltips, context menu, and comboboxes drop shadow

Nuklon:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes

## What is the current behavior?

Currently, the drop shadow used by tooltips and context menus are broken. This is because effects get clipped.
image

## What is the new behavior?

 The solution is to add a decorator that wraps around it. I also updated the dropshadows to follow original colors (they were far too dark). The tooltips have less shadow than context menus in WinUI.

image
image

PR #1516, opened 2025-08-23
fix(controls): `DatePicker` focus

Nuklon:
## Pull request type

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Wpf Ui Gallery_hdSoXXq0GF

This focus looks ugly (accent border goes through the side and it doesn't cover the entire picker), and as DatePicker doesn't seem to have a working FocusVisualStyle, remove the focus style there completely, and allow the user to go to the calendar button with TAB as well (this now DOES focus).

image

I also fixed the missing dropshadow for calender picker:
image

### DIFF
--- a/src/Wpf.Ui/Controls/DatePicker/DatePicker.xaml
+++ b/src/Wpf.Ui/Controls/DatePicker/DatePicker.xaml
@@ -120,7 +120,11 @@
                                 Background="Transparent"
                                 BorderBrush="Transparent"
                                 Cursor="Arrow"
-                                KeyboardNavigation.TabIndex="1">
+                                KeyboardNavigation.TabIndex="1"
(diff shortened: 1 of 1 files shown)


Nuklon:
Depends on https://github.com/lepoco/wpfui/pull/1472.

Nuklon:
With some converters we can still get a proper focus style of the textbox.

image

image

image

Updated the PR. Using TAB still works now, which is nice improvement over the original version.

Nuklon:
We can actually make it much simpler now by just reusing the default TextBox style. This also fixes the darker focus border at the bottom as shown in the above screenshot (which is what I wanted to investigate in the first place).

image

pomianowski (approved):

PR #1534, opened 2025-09-11
fix: Make `NotifyIcon` contextmenu display properly

IOL0ol1:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

NotifyIcon's ContextMenu not display

## What is the new behavior?

- now we can see the NotifyIcon's ContextMenu
-

## Other information

The context menu isn't part of the same visual tree and therefore doesn't have the same DataContext. 
I have provided a tip here to solve this problem.


IOL0ol1:
use x:Name=MainWindow is better, but i want less code.

PR #1547, opened 2025-10-17
fix: `PasswordBox` issues with pasting/copying

Nuklon:
## Pull request type

- [ ] Update
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

There are many issues inside PasswordBox:
1. If you paste a text that's longer than the current text, only the characters at the end are changed. For example. If you have "test" and paste "hello world", it'll become "testo world".
2. If you paste in multiple characters, often only one gets inserted due to breaking early.
3. Selecting multiple characters, and then paste a shorter one, you'll get a mishmash of characters.
4. Exception thrown when binding to a null Password.

This PR addresses all issues by simply using the TextChanges passed in with the event.


Nuklon:
@pomianowski If you have some time, there are a lot of PRs to review/merge.

PR #1569, opened 2025-11-10
fix(win32): `PInvoke` cleanup

Nuklon:
## Pull request type

- [X] Update
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

To avoid cases such as https://github.com/lepoco/wpfui/pull/1566, use CsWin32 and generate these methods from metadata.

## What is the new behavior?

Use CsWin32.

## Other information

Should be no functional changes. Some methods I've changed to check the result from. I also removed a couple of methods from UnsafeNativeMethods as they weren't used anymore (and relied on undocumented code).

### DIFF
--- a/src/Wpf.Ui/Controls/TitleBar/TitleBarButton.cs
+++ b/src/Wpf.Ui/Controls/TitleBar/TitleBarButton.cs
@@ -175,7 +175,7 @@ public class TitleBarButton : Wpf.Ui.Controls.Button
     internal bool ReactToHwndHook(uint msg, IntPtr lParam, out IntPtr returnIntPtr)
     {
         returnIntPtr = IntPtr.Zero;
-        
+
         switch (msg)
         {
             case PInvoke.WM_NCHITTEST:


Nuklon:
@pomianowski We might want to rename UnsafeNativeMethods to PInvokeHelper for example as it's not really native methods. We could also bring the two UnsafeReflection methods into that perhaps.

pomianowski:
> @pomianowski We might want to rename UnsafeNativeMethods to PInvokeHelper for example as it's not really native methods. We could also bring the two UnsafeReflection methods into that perhaps.

Hey @Nuklon. Thanks for your help and huge contribution to the WPF UI. `PInvokeHelper` sounds very generic to me. I stole the name of `UnsafeNativeMethods` from the source code somewhere in .NET SDK. I think that making `Helper` is, as usual, gathering methods into a garbage class that has many responsibilities. I think we can either make direct references to Win32 code, or just something like `static` Dmwapi.GetAccentColor()` 🤷‍♂️

Nuklon:
> > @pomianowski We might want to rename UnsafeNativeMethods to PInvokeHelper for example as it's not really native methods. We could also bring the two UnsafeReflection methods into that perhaps.
> 
> Hey @Nuklon. Thanks for your help and huge contribution to the WPF UI. `PInvokeHelper` sounds very generic to me. I stole the name of `UnsafeNativeMethods` from the source code somewhere in .NET SDK. I think that making `Helper` is, as usual, gathering methods into a garbage class that has many responsibilities. I think we can either make direct references to Win32 code, or just something like `static` Dmwapi.GetAccentColor()` 🤷‍♂️

Glad to help 😊 I think moving some of code inline is fine, but some will be duplicated. I'll give this some thinking first.

Issue #1402, opened 2025-04-01
Please enhance the close button of TabItem


northgisdev:
### Is your feature request related to a problem? Please describe

Please implement a close button for TabItem, just like the tags on Chrome or EDGE browsers.

### Describe the solution you'd like

I have implemented a close button for TabItem myself. If there is a better method, please update the functionality. This is the code I implemented myself. Firstly, add a class. Then update the style. Add Grid and place the close button in the second column.

### Describe alternatives you've considered

public static class TabItemCloseBehavior
{
    public static readonly DependencyProperty IsCloseButtonEnabledProperty =
        DependencyProperty.RegisterAttached(
            "IsCloseButtonEnabled",
            typeof(bool),
            typeof(TabItemCloseBehavior),
            new UIPropertyMetadata(false, OnIsCloseButtonEnabledChanged));

    /// <summary>Helper for getting <see cref="IsCloseButtonEnabledProperty"/> from <paramref name="obj"/>.</summary>
    /// <param name="obj"><see cref="DependencyObject"/> to read <see cref="IsCloseButtonEnabledProperty"/> from.</param>
    /// <returns>IsCloseButtonEnabled property value.</returns>
    [AttachedPropertyBrowsableForType(typeof(DependencyObject))]
    public static bool GetIsCloseButtonEnabled(DependencyObject obj)
    {
        return (bool)obj.GetValue(IsCloseButtonEnabledProperty);
    }

    /// <summary>Helper for setting <see cref="IsCloseButtonEnabledProperty"/> on <paramref name="obj"/>.</summary>
    /// <param name="obj"><see cref="DependencyObject"/> to set <see cref="IsCloseButtonEnabledProperty"/> on.</param>
    /// <param name="value">IsCloseButtonEnabled property value.</param>
    public static void SetIsCloseButtonEnabled(DependencyObject obj, bool value)
    {
        obj.SetValue(IsCloseButtonEnabledProperty, value);
    }

    private static void OnIsCloseButtonEnabledChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
    {
        if (d is TabItem tabItem && e.NewValue is bool newValue && newValue)
        {
            tabItem.Loaded += TabItem_Loaded;
        }
    }

    private static void TabItem_Loaded(object sender, RoutedEventArgs e)
    {
        if (sender is TabItem tabItem)
        {
            // Manually add a close button
            var closeButton = new Wpf.Ui.Controls.Button
            {
                Width = 16,
[...]

Issue #770, opened 2023-09-23
Add "New Tab" & "Close Tab" buttons


RaphMar2021:
### Is your feature request related to a problem? Please describe

I want to make a tabbed app with WpfUi, but there's no option to enable these buttons.

### Describe the solution you'd like

Add a "EnableNavButtons" property in the Wpf.Ui.Controls.TabControl class.

### Describe alternatives you've considered

None

Issue #557, opened 2023-03-12
Add New Tab button & Close tab button


RaphMar2021:
### Is your feature request related to a problem? Please describe

I wanted to make a web browser with WPF UI, but there was no add & close tab button.

### Describe the solution you'd like

Add a New Tab button, and a Close tab button (like in WinUI 3)
![image](https://user-images.githubusercontent.com/81994075/224551045-3a878024-4d41-4e21-a7ac-eafc279c2888.png)

### Describe alternatives you've considered

Nothing

Issue #1038, opened 2024-04-04
WPF UI Gallery shows an alternate `TabControl` when naviagting to it via `Home > Navigation > TabView`


pyroboots:
### Describe the bug

There appear to be 2 different versions of the `TabControl` / `TabView`. `TabView` seems to be a page with an unstyled `TabControl` on and isn't accessible via the sidebar / `NavigationView`. You can only access it by starting on the home page of the app, clicking on "Navigation" and then "TabView".

### To Reproduce

You can only access it by starting on the home page of the app, clicking on "Navigation" and then "TabView".

### Expected behavior

To be taken to the `TabControl` page.

### Screenshots

![image](https://github.com/lepoco/wpfui/assets/118924562/80f37fa6-ecca-41cd-a6d4-c6fb9f9d510b)
![image](https://github.com/lepoco/wpfui/assets/118924562/24fa476e-a120-47a4-a4f0-2898cece5b1d)

### OS version

Windows 11 (22H2), OS Build 22621.3296

### .NET version

Couldn't find it, but it's the .NET version that WPF UI Gallery is running on as of the 4th of April, 2024.

### WPF-UI NuGet version

I'm assuming it's the same version that WPF UI Gallery is running on as of the 4th of April, 2024.

chucker:
Still open as of 153baadeaf475875a81a4605881ca5c2dbe53972, because both `TabViewPage` and `TabControlPage` have a `[GalleryPage]` attribute.

Related to #1587 and others, IMHO: we need to either clarify the distinction of `TabView` and `TabControl`, or merge them.

PR #1587, opened 2025-11-28
feat(controls): Add an “Add Button” and a “Close Button” to the tab control. Additionally, drag-and-drop functionality has been enabled.

Koichi-Kobayashi:
## Pull request type

- [ ] Update
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #557 #770 #1402

## What is the new behavior?

-Added an “Add Button” and a “Close Button”
-Enabled tab dragging and dropping

## Other information
2025-11-29_06h29_47

2025-11-29_06h30_54

https://github.com/user-attachments/assets/4eb16edb-b8b6-46ae-aa2f-90240af54183


Koichi-Kobayashi:
There was an issue where the close button was displayed even when tabs were set not to be closable.

Koichi-Kobayashi:
We have adjusted the display state and spacing of the close button on the tab control. Please review.

chucker:
> Please review.

So, to align this with WinUI 3 where possible, I'm wondering if we should split the control in two:

- the thing _you're_ writing is called the Tab**View** in WinUI 3 Gallery:

image

It has close buttons, drag & drop, etc. Its description says "provides the user with a collection of tabs that can be used to display several documents".

- but the thing a WPF Tab**Control** usually does is IMHO distinct. It's closer to what WinUI 3 Gallery calls a **SelectorBar**

image

…and WinUI 2(?) called it a **Pivot**:

image

_My_ thinking is that low-level, we can do these changes, but high-level, using TabControl _directly_ through WPF-UI should be discouraged; rather, we provide two high-level controls

- SelectorBar
- TabView

…that internally use TabControl, but externally expose the appropriate properties.

Or maybe we can have a property `Behavior` where you can choose `SelectorBar` or `TabView`?

chucker:
See also #847, #1038

src/Wpf.Ui/Controls/TabView/TabView.cs
```diff
@@ -11,4 +18,230 @@ namespace Wpf.Ui.Controls;
...
+        {
+            newTab.Header = header;
+        }
+
+        Items.Add(newTab);
+        SetCurrentValue(SelectedItemProperty, newTab);
+        SetupTabItem(newTab);
+    }
+
+    public override void OnApplyTemplate()
```
sjelatex, line 231: I believe this would cause memory leaks since the previous addButton is never unsubscribed from. A fix would be:

```csharp
public const string PART_AddButton = "PART_AddButton";

private System.Windows.Controls.Button addButton;

public override void OnApplyTemplate()
{
    base.OnApplyTemplate();

    if (addButton != null)
    {
        addButton.Click -= OnAddButtonClick;
    }

    addButton= GetTemplateChild(PART_AddButton) as System.Windows.Controls.Button;

    if (addButton != null)
    {
        addButton.Click += OnAddButtonClick;
    }
}
```

PR #1590, opened 2025-12-03
fix(controls): fix accelerator conflicts and button focus in ContentDialog

### LINKED ISSUE #1585: Close button of simple ContentDialog does not react to Enter key
### Describe the bug

With WPF.UI version 4.1.0 there is a new bug concerning content dialogs. It is likely caused by #1565.

With version 4.0.3 the following behavior occurs when calling simple `ContentDialog` instances via the `IContentDialogService`: if only a `CloseButtonText` was provided via the `SimpleContentDialogCreateOptions`, but no `PrimaryButtonText` or `SecondaryButtonText`, then the content dialog could subsequently be closed by clicking the Close button or pressing the Escape or the Enter key.

Now, with version 4.1.0, the latter no longer works. The Enter key is non-functional.

I suspect this is because the Primary Button is probably the default ("IsDefault") button of the content dialog. And since #1565 all buttons that are not displayed in the content dialog are removed from focus entirely. Thus, pressing the Enter key in a content dialog without Primary Button does not reach the "default" button anymore, and the dialog remains open.

However, if no button other than the Close button (which is the only button required to set!) is displayed, the "IsDefault" characteristic should point to the Close button (instead of the Primary button) so that the dialog can still be closed using the Enter key.

### To Reproduce

In WPF.UI 4.0.3 a content dialog created via the following call can be closed via the Enter key:

```csharp
// _contentDialogService is an instance of IContentDialogService wired to (a singleton) instance of the
// ContentDialogService class.

_ = await _contentDialogService.ShowSimpleDialogAsync(
    new SimpleContentDialogCreateOptions()
    {
        Title = "Dialog title",
        Content = "Dialog content",
        CloseButtonText = "Close"
    });
```

This does not work anymore in WPF.UI 4.1.0.

### Expected behavior

A (simple) content dialog containing only a Close button can be closed using the Enter key.

If this must not be the default behavior for whatever reason, there should be a dependency property (or something else) to wire this manually.

### Screenshots

_No response_

### OS version

Windows 11 (25H2)

### .NET version

.NET 9

### WPF-UI NuGet version

4.1.0

### Additional context

_No response_

apachezy:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- The `ContentDialog` is a UI element that resides within the host window, not an independent window. Its default and cancel buttons are managed by the host window. If both the `ContentDialog` and the host window have default or cancel buttons, the corresponding key presses may be triggered incorrectly (violating the principle that a window should only have one default and one cancel button).

- When a `ContentDialog` contains only a close button, users cannot dismiss it by pressing the Enter key. This violates the muscle memory of most Windows users and deviates from the standard behavior of Windows dialog boxes.

- When the default button (`IsDefault=true`) of a `ContentDialog` does not have keyboard focus, users cannot activate it by pressing the Space key. This breaks the expected keyboard interaction pattern for default buttons.

Issue Number: #1585

## What is the new behavior?

- Registers a temporary AccessKey handler before displaying the content dialog, ensuring that all access keys (including Enter and Esc) from the host window do not interfere with the dialog's keyboard operations.

- When only a close button is available, the `ContentDialog` automatically applies the `IsDefault` property to this button upon loading. This enables dismissal via the Enter key, matching user expectations for single-button dialogs.

- During loading, the `ContentDialog` correctly assigns keyboard focus to the default button (when available). This ensures the button can be activated using the Space key, following standard Windows dialog interaction patterns.

- Removes the previously registered AccessKey handler after the content dialog is closed, thereby restoring the host window's original access key functionality.

- Fixed #1585

## Other information

If you're confused about the first point "erroneous triggering", try placing some buttons in the main window's `TitleBar.TrailingContent`, for example:

```xaml
<ui:TitleBar.TrailingContent>
    <Grid>
        <Grid.ColumnDefinitions>
            <ColumnDefinition Width="Auto" />
            <ColumnDefinition Width="Auto" />
        </Grid.ColumnDefinitions>
        <Button
            Grid.Column="0"
            Click="TestButton_Click"
            Content="Cancel button in TitleBar"
            IsCancel="True" />
        <Button
            Grid.Column="1"
            Margin="8,0"
            Click="TestButton_Click"
            Content="Default button in TitleBar"
            IsDefault="True" />
    </Grid>
</ui:TitleBar.TrailingContent>
```

And implement the click event handler:

```csharp
private void TestButton_Click(object sender, RoutedEventArgs e)
{
    if (sender is System.Windows.Controls.Button button)
    {
        System.Diagnostics.Debug.WriteLine($"TestButton_Click: {button.Content}");
    }
}
```

Then run in debug mode and open a `ContentDialog`. You'll notice these buttons still respond to Enter and Esc keys, which should not be allowed while a modal dialog is open.

### DIFF
--- a/src/Wpf.Ui/Controls/ContentDialog/ContentDialog.cs
+++ b/src/Wpf.Ui/Controls/ContentDialog/ContentDialog.cs
@@ -3,6 +3,8 @@
 // Copyright (C) Leszek Pomianowski and WPF UI Contributors.
 // All Rights Reserved.
 
+using System.Windows.Automation;
+using System.Windows.Automation.Peers;
 using System.Windows.Controls;
 using System.Windows.Input;
 using System.Windows.Threading;
@@ -783,6 +785,12 @@ public class ContentDialog : ContentControl
 
             // Register an AccessKey handler to suppress all access keys of the host window while this dialog is displayed.
             AccessKeyManager.AddAccessKeyPressedHandler(window, HostAccessKeySuppressHandler);
+
+            // Also suppress command execution and can-execute routing from the host window so that
+            // input bindings / command gestures registered on the window won't trigger host commands
+            // while the dialog is shown. Handlers are removed in RestoreHostWindowAccessKeys.
+            CommandManager.AddPreviewExecutedHandler(window, PreviewExecutedSuppressHandler);
+            CommandManager.AddPreviewCanExecuteHandler(window, PreviewCanExecuteSuppressHandler);
         }
         catch
         {
@@ -798,6 +806,8 @@ public class ContentDialog : ContentControl
             if (window != null)
             {
                 AccessKeyManager.RemoveAccessKeyPressedHandler(window, HostAccessKeySuppressHandler);
+                CommandManager.RemovePreviewExecutedHandler(window, PreviewExecutedSuppressHandler);
+                CommandManager.RemovePreviewCanExecuteHandler(window, PreviewCanExecuteSuppressHandler);
             }
         }
         catch
@@ -884,6 +894,39 @@ public class ContentDialog : ContentControl
 
             // Otherwise suppress the access key for this window by clearing the Target.
             e.Target = null;
+        }
+        catch
+        {
+            // ignore
+        }
+    }
+
+    private void PreviewExecutedSuppressHandler(object sender, ExecutedRoutedEventArgs e)
+    {
+        try
+        {
+            if (e.OriginalSource is DependencyObject d && !IsPartOfThisDialog(d))
+            {
+                // Prevent host-window command execution
+                e.Handled = true;
+            }
+        }
+        catch
+        {
+            // ignore
+        }
+    }
+
+    private void PreviewCanExecuteSuppressHandler(object sender, CanExecuteRoutedEventArgs e)
+    {
+        try
+        {
+            if (e.OriginalSource is DependencyObject d && !IsPartOfThisDialog(d))
+            {
+                // Prevent host-window command from being considered executable
+                e.CanExecute = false;
+                e.Handled = true;
+            }
         }
         catch
         {


apachezy:
During this investigation, I also discovered an interesting "problem demonstration": In the current Gallery project, the main window's `TitleBar` is placed behind the `RootContentDialog` (the `TitleBar` has a higher Z-order than the `ContentDialog`). As a result, if the toolbar contains some controls, they are **not blocked from interaction** when a ContentDialog is open. For example:

*(example image 1)*
error1

Controls in the toolbar can still receive focus and clicks...

After multiple tests, I believe this should indeed be considered an issue: after interacting with these unblocked controls, focus returns to the main window, and users can even Tab-navigate to other UI elements in the main window:

*(example image 2)*
error2

Issue Resolved: Simply place `RootContentDialog(ContentPresenter)` at the topmost layer in XAML (i.e., at the bottom in the XAML markup order). For example:

```xaml
<ui:FluentWindow>
    <Grid>
        <ui:NavigationView>
            ...
        </ui:NavigationView>

        <ui:TitleBar>
            ...
        </ui:TitleBar>

        <!--  Place it here  -->
        <ContentPresenter x:Name="RootContentDialog" Grid.Row="0" />
    </Grid>
</ui:FluentWindow>
```

This does not affect the interactive behavior of `TitleBar` itself, as it uses a special mechanism to handle interaction.

apachezy:
Regarding Automation/Accessibility

Assistive tools such as Windows Narrator can still read content from the host window and move input focus away from the ContentDialog. I have not found a suitable solution for this issue.

Although I could disable all UI elements at the same level as `RootContentDialog` (`IsEnabled = "False"`) before showing the `ContentDialog` in my own projects, this does not seem to be the best solution, as the appearance of elements changes after being disabled.

I really like `ContentDialog` and have used it in many projects to replace `MessageBox`, as well as for forms, dialogs, and more.

I hope experts can solve this problem...

apachezy:
I think I've finally nailed the accessibility focus escape issue for ContentDialog. Closing this PR for now—will be back with a shiny new PR once everything's tested and polished. ✨

Koichi-Kobayashi:
@apachezy 
There's also an issue where focus is lost in MessageBox, and I've submitted a pull request for #1595.
If we're fixing it using a similar method, it seems feasible, but would a different approach be better?

apachezy:
Hello @Koichi-Kobayashi ,  

Thank you for your trust and for reaching out. I appreciate your initiative in improving the message box.  

Regarding your current patch, here are a few areas that could be further refined:  

1. **Calling `SetFocusToFirstAvailableButton` in the `Loaded` event without a delay:**  
   Since the `Loaded` event already ensures an appropriate timing for setting focus, adding a delay here may interfere with the focus logic of other developers who might implement their own focus-handling routines.  

2.  **Check if the `MessageBox` itself currently has focus before setting focus:**
    Before setting focus programmatically, we should first verify whether `MessageBox.IsFocused` is `true`. If it's `false`, it indicates that focus is already on one of the internal controls (e.g., a `TextBox`) within the message box. This likely means a developer's focus logic has already taken effect, and we should avoid overriding it.

3. **Refine `SetFocusToFirstAvailableButton` to avoid searching the entire visual tree unnecessarily:**  
   Currently, `FindButtonsInVisualTree` searches for default buttons across the entire visual tree, which can be expensive and may interfere with developers who have customized the message box template with their own default buttons. Instead, we should focus only on the three default buttons of the message box. For example, we can assign names to these buttons and retrieve them using `GetTemplateChild`, which is more efficient and targeted.  

These are just my thoughts based on the current implementation. The design of focus behavior can be nuanced, and I might not have considered all scenarios. I'm open to further discussion if you have different ideas or insights on this.

PR #1519, opened 2025-08-24
Allowed users to change the width of NavigationView with the mouse.

Koichi-Kobayashi:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Update
- [ ] Bugfix
- [ x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

-I added transparent lines.
2025-08-25_05h52_04
2025-08-25_05h53_45

PR #1603, opened 2025-12-13
feat(controls): Add GridSplitter support to LeftNavigationViewTemplate

Koichi-Kobayashi:
## Pull request type

- [ ] Update
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The width of the navigation panel can be specified using OpenPaneLength,
but it cannot be resized dynamically.

I came up with a better approach than my pull request #1519.

## What is the new behavior?

The GridSplitter allows users to manually resize the navigation pane width,
and the navigation menu automatically adjusts its width accordingly while
staying left-aligned. The initial width is controlled by the OpenPaneLength property.

- Add IsGridSplitterEnabled dependency property to enable/disable GridSplitter
- Implement GridSplitter between navigation pane and content area
- Ensure navigation pane width resizes with GridSplitter while maintaining left alignment
- Manage PaneColumn width through code-behind using OpenPaneLength as initial width
- Remove ColumnDefinition width animations from VisualStateManager to prevent exceptions
- Add proper event handler cleanup and thread-safe updates using Dispatcher
- Handle GridLength safely with IsAbsolute checks before accessing Value property
- Default value is false to maintain backward compatibility

## Other information

https://github.com/user-attachments/assets/ed6a467d-738e-45ae-8555-5f9f356e0d5d


Nuklon:
This destroys the navigation animations and makes the search box collapse extremely ugly.

Koichi-Kobayashi:
@Nuklon 

Sorry about that.
I previously posted the following, but the video was incorrect. I’ll double-check it. (It seems the existing Gallery was on the right.)

> I recorded a video of the part where the navigation seems to be broken.
> 
> The version overlaid on the right is the one I modified, and I noticed the following two issues.
> Could you please confirm whether my understanding is correct?
> 
> - When clicking the hamburger menu to close the navigation, it shifts slightly upward.
> - The open/close animation of the navigation has become slower.

https://github.com/user-attachments/assets/f436a375-0ee1-4305-a9ff-c8fd84795a7f

Issue #972, opened 2024-02-25
FluentWindow not allowing window resize from left or right top corners


Roypy:
### Describe the bug

When trying to resize the window from either the top left corner or the top right corner the resize cursor does not appear.

### To Reproduce

1. Setup WpfUi
2. Add a titlebar
3. Try to resize from upper corners

### Expected behavior

The windows should be resizable from every corner, not just the bottom left and bottom right

### Screenshots

![Error](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExY21ydTA0b2dxajV2ZHNvbGoyajAwb2x6ZGN3ejJrNHBlMm80ZDM5dyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/dfKA6oyOePxTHnkpiD/giphy.gif)

### OS version

Window 11 22H2 (OS Build 22621.3155)

### .NET version

.NET 8.0

### WPF-UI NuGet version

3.0.0

seasonyuu:
This is a bug since v3.0.0, v2.x works like a charm

Paliverse:
When giving a small margin to the root element in the Fluent Window, it will work in all corners and sides.

![image](https://github.com/lepoco/wpfui/assets/4289084/a0d7b9a3-2252-4648-9360-c40a160181fd)

PR #1604, opened 2025-12-13
fix(controls): Enable window resizing from title bar corners

Koichi-Kobayashi:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

I was unable to resize the window from the top-left and top-right corners.

Issue Number: #972

## What is the new behavior?

This enables users to resize windows by dragging from the top-left and
top-right corners of the title bar, even when title bar buttons are present.

- Set WindowChrome ResizeBorderThickness top value to title bar height to allow resizing from the title bar area
- Add priority resize detection for 10px square areas at top-left and top-right corners
- Skip button hit testing when resize detection succeeds at corners to ensure proper resize behavior

## Other information

https://github.com/user-attachments/assets/01856b26-f631-4303-a64a-b3826cd8fdf1


Nuklon:
The caption height shouldn't be used, it's 22px for me, so you'll have massive resize area (22px) on top and bottom. Same with cornerResizeSize, which should also be 4px.

Another downside is that if you use TitleBar.TrailingContent with some buttons (fairly common) then the resizer will become active when you hover over buttons.

Koichi-Kobayashi:
@Nuklon 
I addressed the feedback about the resize hit-test area being too large and interfering with interactive title-bar content.

- Corner resize area is now fixed to 4px: In TitleBar.WindowResize.cs the top-left/top-right corner hit-test (WM_NCHITTEST) uses a 4px square instead of the previous 10px, so the resize cursor only appears in the intended small corner region.
- Do not use CaptionHeight for resize border: In FluentWindow.SetWindowChrome() I removed the SystemParameters.CaptionHeight usage and set WindowChrome.ResizeBorderThickness to a predictable new Thickness(4, 4, 4, 4), avoiding a massive top resize area on systems where caption height is ~22px.
- Prevent resizing over custom title-bar content (e.g. TrailingContent buttons): GetWindowBorderHitTestResult now receives a flag indicating whether the pointer is over interactive title-bar content (Header/CenterContent/TrailingContent). If it is, the method returns HTNOWHERE (except for the 4px corner zones), so hovering buttons in TrailingContent no longer activates the resizer.

I also merged main into the branch and resolved conflicts by keeping main’s expanded “interactive content” detection (including CenterContent) while preserving the new hit-test behavior.

Koichi-Kobayashi:
There was an unresolved issue, so I'm opening a new pull request.

PR #1626, opened 2025-12-27
fix(controls): Resolve TextBlock style improperly affecting global scope

### LINKED ISSUE #1625: Issue with TextBlock.xaml: problematic implicit global style and suggestions
**Description:**

Regarding the current `TextBlock.xaml`, which defines an implicit global style for `TextBlock`, I believe it contains several meaningless initial settings and even harmful default property assignments. It should be either removed or significantly revised.

**Current Problems:**

1. **Performance Impact**
  `Background="Transparent"` – This forces rendering of transparent backgrounds across all TextBlocks. For large-scale usage, this impacts performance. It should remain `null` (default) unless explicitly needed.

2. **Unnecessary Explicit Defaults**  
  `Margin`, `Padding`, `Focusable` – These properties are already set to their default values by WPF. Explicitly assigning them in a global implicit style is redundant and adds unnecessary complexity. Global implicit styles should remain minimal and avoid redundant setters that offer no functional benefit.

3. **Over‑specified Rendering**
  `SnapsToDevicePixels` – Should be controlled by developers or inherited from parent containers, not forced globally.

4. **Misuse of `OverridesDefaultStyle`**  
  `OverridesDefaultStyle="True"` – This property is **meaningless and harmful** in an implicit global style. Its correct usage is in the default style of a **custom control** (typically applied within the static constructor of the control class), not in an application‑level implicit style. Applying it here bypasses the WPF theme system in unintended ways and can lead to inconsistent behavior.

5. **Redundant/Theming Conflict**
  `FontSize` – This pollutes the global `TextBlock` font size. When using `ui:TextBlock`, the default `FontTypography` property already provides appropriate sizing.

**Impact:**
This implicit style affects **all WPF TextBlocks** in applications that reference WPFUI, not just `ui:TextBlock`. Some of these forced defaults can cause unexpected behavior, performance degradation, and layout issues.

**Suggested Solutions (choose one):**

1. **Remove the file entirely** – Let `ui:TextBlock` rely on its own logic and typography system, and leave standard `TextBlock` completely untouched by WPFUI.

2. **Revise the style to be WPFUI‑specific** – Change the style to target only `ui:TextBlock` (if architecturally possible) and remove all harmful or unnecessary setters.

3. **Keep only essential theming properties** – Retain only `FontSize` and `Foreground="{DynamicResource TextFillColorPrimaryBrush}"`. This minimal set ensures that standard WPF `TextBlock` elements follow the same theme behavior as `ui:TextBlock`, maintaining visual consistency across the application without introducing unnecessary overrides.

apachezy:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

As documented in issue #1625 , the implicit global style in `TextBlock.xaml` affects **all WPF TextBlock** controls in applications using WPFUI. This introduces unnecessary and potentially harmful defaults (`OverridesDefaultStyle="True"`, forced `Background="Transparent"`, redundant `Margin`/`Padding`, etc.), which can cause unexpected behavior, performance impact, and conflicts with standard WPF styling.

Issue Number: #1625

## What is the new behavior?

- **File removed** – `TextBlock.xaml` has been deleted entirely.

- **WPF TextBlock remains unaffected** – Standard WPF `TextBlock` controls are no longer modified by WPFUI, preserving their default theme behavior.

- **No impact on ui:TextBlock** – The WPFUI-enhanced `ui:TextBlock` continues to work correctly with its own typography and styling logic.

- **Cleaner and safer** – All incorrectly applied properties and global overrides are eliminated, removing associated risks and maintaining better performance and styling semantics.

- Fixed #1625


apachezy:
After further consideration, I believe the optimized approach (refining rather than removing) is more aligned with WPFUI's design philosophy and provides better value to developers. I'll submit an alternative PR that keeps minimal theming while removing all problematic defaults.

PR #1627, opened 2025-12-27
fix(controls): Refine TextBlock global style to remove harmful defaults

### LINKED ISSUE #1625: Issue with TextBlock.xaml: problematic implicit global style and suggestions

apachezy:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

As documented in issue #1625, the implicit global style in `TextBlock.xaml` affects all WPF TextBlock controls in applications using WPFUI. While providing basic theming consistency, it introduces problematic defaults:

- `OverridesDefaultStyle="True"` (incorrect usage in global implicit style)

- `Background="Transparent"` (forces unnecessary rendering overhead)

- Redundant `Margin`, `Padding`, `Focusable` setters

- `SnapsToDevicePixels` (should be context-dependent, not globally forced)

Issue Number: #1625

## What is the new behavior?

- **Harmful defaults removed** – All problematic properties listed above are eliminated

- **Minimal theming retained** – Only essential theme properties remain:
  
  - `FontSize` (ensures consistent typography scaling)

  - `Foreground="{DynamicResource TextFillColorPrimaryBrush}"` (theme-aware text color)

- **Clean theme integration** – Standard WPF `TextBlock` now automatically follows WPFUI's theme system, reducing developer styling boilerplate

- **No breaking changes** – `ui:TextBlock` continues to work as expected, while native controls gain safe theme consistency

- **Aligned with existing patterns** – Consistent with how other WPFUI controls (e.g., TextBox) apply base theming to native WPF counterparts

- Fixed #1625

## Other information

**Why this approach:**

- Fixes the actual issues without throwing away useful theming

- Maintains backward compatibility and developer experience

- Follows the library's established design philosophy

**Note on the Commented-Out Foreground Setter**

I noticed the Foreground setter was commented out while FontSize remained active. This creates an inconsistent theming approach:

- If we're theming TextBlock at all, both typography and color should follow WPFUI's design system

- The current "half-theming" provides limited value while still affecting global behavior

- TextFillColorPrimaryBrush is a core part of WPFUI's theme engine — enabling it gives developers automatic light/dark mode support

This PR uncomments and enables the Foreground setter, completing the minimal theming vision that was apparently originally intended.

### DIFF
--- a/src/Wpf.Ui.Gallery/Views/Pages/BasicInput/ButtonPage.xaml
+++ b/src/Wpf.Ui.Gallery/Views/Pages/BasicInput/ButtonPage.xaml
@@ -71,10 +71,14 @@
                     <ColumnDefinition Width="*" />
                     <ColumnDefinition Width="Auto" />
                 </Grid.ColumnDefinitions>
-                <ui:Button
+                <Grid.Resources>
+                    <Style TargetType="ui:TextBlock" BasedOn="{StaticResource {x:Type ui:TextBlock}}">
+                        <Setter Property="Foreground" Value="BlueViolet"></Setter>
+                    </Style>
+                </Grid.Resources>
+                <ui:TextBlock
                     Grid.Column="0"
-                    Content="WPF UI button"
-                    Icon="{ui:SymbolIcon Fluent24}"
+                    Text="WPF UI button"
                     IsEnabled="{Binding ViewModel.IsUiButtonEnabled, RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=local:ButtonPage}, Mode=OneWay}" />
                 <CheckBox
                     Grid.Column="1"
--- a/src/Wpf.Ui.Gallery/Views/Pages/DesignGuidance/ColorsPage.xaml
+++ b/src/Wpf.Ui.Gallery/Views/Pages/DesignGuidance/ColorsPage.xaml
@@ -34,12 +34,19 @@
             Padding="12"
             Background="{ui:ThemeResource ControlFillColorDefaultBrush}"
             CornerRadius="8">
+
+            <Border.Resources>
+                <Style TargetType="ui:TextBlock">
+                    <Setter Property="Foreground" Value="ForestGreen" />
+                </Style>
+            </Border.Resources>
             <StackPanel>
                 <ui:TextBlock Text="For UI labels and static text" />
                 <ui:TextBlock
                     Margin="0,12,0,24"
                     HorizontalAlignment="Center"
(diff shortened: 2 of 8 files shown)


apachezy:
This PR will be closed and will be fixed along with other issues in a separate PR.

PR #1629, opened 2025-12-30
fix(controls): Fix Separator layout and add Orientation attached property

apachezy:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, the Separator control uses a global implicit style to apply WPFUI’s theme colors and attempts to support vertical orientation. However, several issues cause incorrect layout behavior:

- **"Double Margin" Effect**: The implicit style overrides the default template and incorrectly makes the internal Border's Margin property inherit the Separator control's Margin property via TemplateBinding.

  As shown below:
  ```xml
  <Style TargetType="{x:Type Separator}">
      ...
      <Setter Property="Template">
          <Setter.Value>
              <ControlTemplate TargetType="{x:Type Separator}">
                  <!-- Incorrectly inherits the Margin property -->
                  <Border
                      Margin="{TemplateBinding Margin}"
                      ... />
              </ControlTemplate>
          </Setter.Value>
      </Setter>
  </Style>
  ```
  This causes the Separator's own Margin property to take effect, and the internal Border also inherits the same Margin value. The Border then creates spacing inside the control, expanding it and ultimately manifesting as a "double margin" effect.

  Example:
  Separator-Bug

- **Directional Support**: It uses the Top and Left borders of a Border to simulate horizontal and vertical separator lines. When width or height is set larger than 1, two lines become visible. Moreover, these lines are not centered—unlike the native WPF Separator, whose line always stays in the middle.

  Example: `<Separator Width="100" Height="100" />`
  Separator-Border

- **Property and Resource Key Naming**: Due to the above design (using Border to draw the line), the style uses `BorderBrush` for theming, and the resource key is named “SeparatorBorderBrush.” However, the native WPF Separator uses the `Background` property for its visual color. This discrepancy can confuse developers and lead to unexpected behavior.

Issue Number: N/A

## What is the new behavior?

- The style has been redefined to align with the behavior of the native WPF Separator control, eliminating the “double margin” issue.

- Introduced the `SeparatorAttached` helper class, which adds an `Orientation` attached property to Separator. Together with the updated style, this enables seamless direction switching.

- Renamed the theme resource key from **SeparatorBorderBrush** to **SeparatorBackgroundBrush**, matching WPF’s native naming pattern and unifying developer expectations.

Examples:
```xaml
<Separator Margin="0 10 0 2" />
```
Separator-Fixed

and:

```xaml
<Separator ui:SeparatorAttached.Orientation="Vertical" Margin="10" />
```
Separator-Ver

### DIFF
--- a/src/Wpf.Ui/Controls/Separator/Separator.xaml
+++ b/src/Wpf.Ui/Controls/Separator/Separator.xaml
@@ -5,28 +5,47 @@
     All Rights Reserved.
 -->
 
-<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:controls="clr-namespace:Wpf.Ui.Controls">
 
-    <Style TargetType="{x:Type Separator}">
-        <Setter Property="BorderBrush" Value="{DynamicResource SeparatorBorderBrush}" />
-        <Setter Property="Background" Value="Transparent" />
-        <Setter Property="Margin" Value="0" />
-        <Setter Property="BorderThickness" Value="1,1,0,0" />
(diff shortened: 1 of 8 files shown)


apachezy:
This PR got messed up with my commits, so I'm going to close it.

PR #1630, opened 2025-12-31
fix(controls): Fix Separator layout and add Orientation attached prop…

apachezy:
## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, the Separator control uses a global implicit style to apply WPFUI’s theme colors and attempts to support vertical orientation. However, several issues cause incorrect layout behavior:

- **"Double Margin" Effect**: The implicit style overrides the default template and incorrectly makes the internal Border's Margin property inherit the Separator control's Margin property via TemplateBinding.

  As shown below:
  ```xml
  <Style TargetType="{x:Type Separator}">
      ...
      <Setter Property="Template">
          <Setter.Value>
              <ControlTemplate TargetType="{x:Type Separator}">
                  <!-- Incorrectly inherits the Margin property -->
                  <Border
                      Margin="{TemplateBinding Margin}"
                      ... />
              </ControlTemplate>
          </Setter.Value>
      </Setter>
  </Style>
  ```
  This causes the Separator's own Margin property to take effect, and the internal Border also inherits the same Margin value. The Border then creates spacing inside the control, expanding it and ultimately manifesting as a "double margin" effect.

  Example:
  Separator-Bug

- **Directional Support**: It uses the Top and Left borders of a Border to simulate horizontal and vertical separator lines. When width or height is set larger than 1, two lines become visible. Moreover, these lines are not centered—unlike the native WPF Separator, whose line always stays in the middle.

  Example: `<Separator Width="100" Height="100" />`
  Separator-Border

- **Property and Resource Key Naming**: Due to the above design (using Border to draw the line), the style uses `BorderBrush` for theming, and the resource key is named “SeparatorBorderBrush.” However, the native WPF Separator uses the `Background` property for its visual color. This discrepancy can confuse developers and lead to unexpected behavior.

Issue Number: N/A

## What is the new behavior?

- The style has been redefined to align with the behavior of the native WPF Separator control, eliminating the “double margin” issue.

- Introduced the `SeparatorAttached` helper class, which adds an `Orientation` attached property to Separator. Together with the updated style, this enables seamless direction switching.

- Renamed the theme resource key from **SeparatorBorderBrush** to **SeparatorBackgroundBrush**, matching WPF’s native naming pattern and unifying developer expectations.

Examples:
```xaml
<Separator Margin="0 10 0 2" />
```
Separator-Fixed

and:

```xaml
<Separator ui:SeparatorAttached.Orientation="Vertical" Margin="10" />
```
Separator-Ver

### DIFF
--- a/src/Wpf.Ui/Controls/Separator/SeparatorAttached.cs
+++ b/src/Wpf.Ui/Controls/Separator/SeparatorAttached.cs
@@ -1,0 +1,91 @@
+// This Source Code Form is subject to the terms of the MIT License.
+// If a copy of the MIT was not distributed with this file, You can obtain one at https://opensource.org/licenses/MIT.
+// Copyright (C) Leszek Pomianowski and WPF UI Contributors.
+// All Rights Reserved.
+
+using System.Windows.Controls;
+
+// ReSharper disable once CheckNamespace
+namespace Wpf.Ui.Controls;
+
+/// <summary>
+/// Provides <c>Orientation</c> property for Separator
+/// </summary>
+/// <example>
+/// <code lang="xml">
+/// &lt;Separator controls:SeparatorAttached.Orientation="Vertical" /&gt;
+/// </code>
+/// </example>
+public static class SeparatorAttached
+{
+    /// <summary>
+    /// Resource key for <see cref="Separator"/> horizontal margin.
+    /// </summary>
+    public const string HorizontalMarginKey = "DefaultSeparatorHorizontalMargin";
+
+    /// <summary>
+    /// Resource key for <see cref="Separator"/> vertical margin.
+    /// </summary>
+    public const string VerticalMarginKey = "DefaultSeparatorVerticalMargin";
+
+    public static readonly DependencyProperty OrientationProperty =
+        DependencyProperty.RegisterAttached(
+            "Orientation",
+            typeof(Orientation),
+            typeof(SeparatorAttached),
+            new FrameworkPropertyMetadata(Orientation.Horizontal, OnOrientationChanged)
+        );
+
+    /// <summary>Helper for getting <see cref="OrientationProperty"/> from <paramref name="obj"/>.</summary>
+    /// <param name="obj"><see cref="DependencyObject"/> to read <see cref="OrientationProperty"/> from.</param>
+    /// <returns>Orientation property value.</returns>
+    [AttachedPropertyBrowsableForType(typeof(Separator))]
+    public static Orientation GetOrientation(DependencyObject obj)
+    {
+        return (Orientation)obj.GetValue(OrientationProperty);
+    }
+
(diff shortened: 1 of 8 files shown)


src/Wpf.Ui/Controls/Separator/SeparatorAttached.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// <summary>Helper for setting <see cref="OrientationProperty"/> on <paramref name="obj"/>.</summary>
+    /// <param name="obj"><see cref="DependencyObject"/> to set <see cref="OrientationProperty"/> on.</param>
+    /// <param name="value">Orientation property value.</param>
+    public static void SetOrientation(DependencyObject obj, Orientation value)
+    {
+        obj.SetValue(OrientationProperty, value);
+    }
+
+    private static void OnOrientationChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
```
Nuklon: This is not needed, you can simply change the margin in xaml.

```

                        <Trigger Property="controls:SeparatorAttached.Orientation" Value="Vertical">
                            <Setter TargetName="SeparatorBorder" Property="Width" Value="1" />
                            <Setter TargetName="SeparatorBorder" Property="Height" Value="Auto" />
                            <Setter TargetName="SeparatorBorder" Property="HorizontalAlignment" Value="Center" />
                            <Setter TargetName="SeparatorBorder" Property="VerticalAlignment" Value="Stretch" />
                            <Setter Property="Margin" Value="{DynamicResource DefaultSeparatorVerticalMargin}" />
                        </Trigger>
```

src/Wpf.Ui/Controls/Separator/SeparatorAttached.cs
```diff
@@ -0,0 +1,91 @@
...
+
+/// <summary>
+/// Provides <c>Orientation</c> property for Separator
+/// </summary>
+/// <example>
+/// <code lang="xml">
+/// &lt;Separator controls:SeparatorAttached.Orientation="Vertical" /&gt;
+/// </code>
+/// </example>
+public static class SeparatorAttached
```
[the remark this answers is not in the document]
apachezy: The name **`Separator`** is intentionally reserved for a potential future `ui:Separator` control.

Keeping the attached behavior under a distinct name avoids naming conflicts and allows a smooth transition if a dedicated control is introduced later.

src/Wpf.Ui/Controls/Separator/SeparatorAttached.cs
apachezy: `OnOrientationChanged` contains logic that detects whether the existing `Margin` value is still the default.
If the margin has been explicitly set by the user, it is preserved; otherwise, it is adjusted automatically when the orientation changes.

This kind of conditional logic cannot be reliably implemented in XAML alone, which is why it needs to remain in code.

PR #1706, opened 2026-05-02
fix(tray): support Unicode for NotifyIcon.TooltipText

### LINKED ISSUE #1705: Setting `NotifyIcon.TooltipText` to non-ASCII text renders as `?` characters in the tray tooltip
## Summary

Setting `NotifyIcon.TooltipText` to a string containing non-ASCII characters results in those characters being replaced with `?` in the tray tooltip on systems whose system ANSI codepage cannot represent them. The same defect affects `szInfo` (balloon body) and `szInfoTitle` (balloon title).

## Reproduction

Tested on en-US Windows 11 24H2 (build 26200), WPF-UI 4.2.1, .NET 10:

```xml
<tray:NotifyIcon TooltipText="80% 已连接(未充电)" Icon="..." />
```

Hovering the tray icon shows `80% ???(???)`. The percentage and ASCII parens render correctly; only the CJK characters are lost.

## Root cause

`src/Wpf.Ui.Tray/Interop/Shell32.cs::NOTIFYICONDATA` is declared without an explicit `CharSet`:

```csharp
[StructLayout(LayoutKind.Sequential)]
public class NOTIFYICONDATA
{
    ...
    [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 0x80)]
    public string? szTip;
    ...
}
```

C# defaults `StructLayoutAttribute.CharSet` to `Ansi`. `UnmanagedType.ByValTStr` follows the enclosing struct's charset, so `szTip`/`szInfo`/`szInfoTitle` are marshaled via `Encoding.Default.GetBytes(...)`, which substitutes `?` for code points unrepresentable in the active ANSI codepage. The matching import:

```csharp
[DllImport(Libraries.Shell32)]
public static extern bool Shell_NotifyIcon([In] NIM dwMessage, [In] NOTIFYICONDATA lpdata);
```

also has no `CharSet`, so it resolves to `Shell_NotifyIconA` and Windows stores the already-corrupted text.

## Suggested fix

Two-line change:

```csharp
[StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode)]
public class NOTIFYICONDATA { ... }

[DllImport(Libraries.Shell32, CharSet = CharSet.Unicode)]
public static extern bool Shell_NotifyIcon([In] NIM dwMessage, [In] NOTIFYICONDATA lpdata);
```

`ByValTStr`'s `SizeConst` slots are interpreted per the new charset (i.e. 128 wide chars instead of 128 ANSI bytes for `szTip`), so the struct's binary size grows accordingly. `cbSize = Marshal.SizeOf(typeof(NOTIFYICONDATA))` recomputes correctly and routes the call through `Shell_NotifyIconW`. No call-site changes required.

## Workaround for consumers

Subclass `NotifyIcon`, override `OnPropertyChanged`, and after `base` re-send the tooltip via your own Unicode-charset NOTIFYICONDATA mirror + `Shell_NotifyIconW(NIM.MODIFY, NIF.TIP)`. Reflection on the existing internal `ShellIconData` recovers `hWnd` and `uID`. Re-apply on both `TooltipTextProperty` and `IconProperty` changes — `TrayManager.ModifyIcon` re-sends the same ANSI `szTip` from the cached struct, which would otherwise overwrite the Unicode fix on every icon update.

## Environment

- WPF-UI: 4.2.1
- Target framework: `net10.0-windows10.0.26100.0` (also reproduces on net8.0-windows)
- OS: Windows 11 24H2 (build 26200)
- System ANSI codepage: 1252 (en-US)
- Tooltip text language: zh-Hans

chucker:
Explicitly use the Unicode (wide) variant of `Shell_NotifyIcon`.

## Pull request type

- [ ] Update
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: fixes #1705

## Other information

I tried to migrate this to `[LibraryImport]` and `CsWin32`, but both of those seemed to be too much of a rabbit hole.

### DIFF
--- a/src/Wpf.Ui.Tray/Interop/Shell32.cs
+++ b/src/Wpf.Ui.Tray/Interop/Shell32.cs
@@ -69,7 +69,7 @@ internal static class Shell32
         VISTA_MASK = XP_MASK | REALTIME | SHOWTIP,
     }
 
-    [StructLayout(LayoutKind.Sequential)]
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode)]
     public class NOTIFYICONDATA
     {
         /// <summary>
@@ -153,7 +153,7 @@ internal static class Shell32
         [Out, MarshalAs(UnmanagedType.Interface)] out object ppv
     );
 
-    [DllImport(Libraries.Shell32)]
+    [DllImport(Libraries.Shell32, CharSet = CharSet.Unicode)]
     [return: MarshalAs(UnmanagedType.Bool)]
     public static extern bool Shell_NotifyIcon([In] NIM dwMessage, [In] NOTIFYICONDATA lpdata);
 


pomianowski (approved):

BlyZeDev:
I implemented a NotifyIcon using LibraryImport. Might be of good use here: https://github.com/BlyZeDev/DotTray

chucker:
> I implemented a NotifyIcon using LibraryImport. Might be of good use here: https://github.com/BlyZeDev/DotTray

Thanks. I tried to use LibraryImport or CsWin32, but it seemed non-trivial to make them not clash with PolySharp.

PR #1733, opened 2026-06-12
fix(controls): Button foreground on pressed showing wrong color in dark mode

### LINKED ISSUE #1722: [Bug] After 4.2.1, ui:Button foreground error on pressed
### Describe the bug

After v4.2.1, ui:Button foreground error on pressed.

### To Reproduce

```xaml
<ui:Button Margin="8,0,0,0" Content="Not good" />
<Button Margin="8,0,0,0" Content="OK" />
```

### Expected behavior

ui:Button foreground should remain white when pressed in dark mode.

### Screenshots

Normal:
Image

Pressed:
Image

### OS version

win11

### .NET version

net10.0-windows

### WPF-UI NuGet version

4.2.1, 4.3.0

### Additional context

_No response_

JyslaFancy:
## Description

Fixes #1722

### Root Cause

Commit 4c711b6a (released in v4.2.1) changed the IsPressed MultiTrigger and IsEnabled (Disabled) trigger in the  ControlTemplate from directly setting  on  and  to setting  on the TemplatedParent (Button), relying on  to propagate the value to the ContentPresenter elements.

However, WPF's  does not always properly reflect trigger-set values on the templated parent. When a ControlTemplate trigger sets  on the parent Button, the TemplateBinding on the ContentPresenter may fail to pick up this change, causing the ContentPresenter to use the default  value () instead of the Style-setter value ().

In dark theme, this results in the button text showing the wrong (dark) color when pressed.

### Fix

Restored the direct  setters on  and  in both the IsPressed MultiTrigger and the IsEnabled (Disabled) trigger, matching the working approach from v4.2.0.

### DIFF
--- a/src/Wpf.Ui/Controls/Button/Button.xaml
+++ b/src/Wpf.Ui/Controls/Button/Button.xaml
@@ -460,7 +460,8 @@
                             </MultiTrigger.Conditions>
                             <Setter TargetName="ContentBorder" Property="Background" Value="{Binding PressedBackground, RelativeSource={RelativeSource TemplatedParent}}" />
                             <Setter TargetName="ContentBorder" Property="BorderBrush" Value="{Binding PressedBorderBrush, RelativeSource={RelativeSource TemplatedParent}}" />
-                            <Setter Property="Foreground" Value="{Binding PressedForeground, RelativeSource={RelativeSource TemplatedParent}}" />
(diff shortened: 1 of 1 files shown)


Nuklon (approved):

Nuklon:
Actually dupe of https://github.com/lepoco/wpfui/pull/1723.

Issue #1315, opened 2025-01-14
Wpf.Ui.Controls.NumberBox event ValueChanged called early


APerricone:
### Describe the bug

ValueChanged event is called before the binded value changes

### To Reproduce

with this piece of code we can see that NumberBox.Value is the correct value but the binded value doesn't.
```xml
         <StackPanel Grid.Row="2" Margin="20" Grid.ColumnSpan="2">
            <ui:NumberBox Value="{Binding testValue}" ValueChanged="NumberBox_ValueChanged"  HorizontalAlignment="Left"/>
            <TextBlock x:Name="bugReport" TextWrapping="Wrap"/>
         </StackPanel>
```
```c#
      public double testValue { get; set; } = 10.12345567788;
      private void NumberBox_ValueChanged(object sender, System.Windows.RoutedEventArgs e) {
         if(bugReport == null)
            return;
         bugReport.Text += $" - {( (NumberBox)sender ).Value.ToString()} ({testValue})";
      }
```

### Expected behavior

The event is called after the binded value is associated

### Screenshots

![image](https://github.com/user-attachments/assets/09c168f6-4ce6-463f-bf67-34e37c6f26c5)

### OS version

Window 11

### .NET version

.net 9.0

### WPF-UI NuGet version

3.0.5

p6laris:
I'm experiencing the same issue. However, when I adjust the value using the up and down arrows, it works fine.

I also noticed that the border of the `NumberBox` turns red when I try to set the value for the second time, I'm not entirely sure why this happens, but as @APerricone mentioned, it might be due to a delay in the binding.

Issue #1016, opened 2024-03-25
NumberBox Increment and Decrement Set the Value After Validation


c0nstexpr:
### Describe the bug

Increment and decrement button in NumberBox invoke [StepValue](https://github.com/lepoco/wpfui/blob/main/src/Wpf.Ui/Controls/NumberBox/NumberBox.cs#L414) function.
It validates input before increment or decrement.
```
        // Before adjusting the value, validate the contents of the textbox so we don't override it.
        ValidateInput();

        var newValue = Value ?? 0;

        if (change is not null)
        {
            newValue += change ?? 0d;        // <--------- increment or drcement happens here
        }

        SetCurrentValue(ValueProperty, newValue);        // <--------- value set here
```

### To Reproduce

Use NumberBox with max and min range set. Click increment or decrement button to set the value out of range.

### Expected behavior

Value should not be set out of range

### OS version

Same as repo

### .NET version

Same as repo

### WPF-UI NuGet version

Same as repo

Issue #945, opened 2024-02-09
NumberBox delayed value propagation after keyboard input


KillyMXI:
### Describe the bug

- It takes two times to leave `ui:NumberBox` focus for the input value to propagate into binded ViewModel;
- Changing to `UpdateSourceTrigger=PropertyChanged` makes it behave like is should've behaved with `UpdateSourceTrigger=LostFocus`.

### To Reproduce

I made this reproduction:
https://github.com/KillyMXI/wpf-ui-numberbox-issue-repro

### Expected behavior

- It should not have extra internal propagation step.
- Value and Text should be updated at the same time.
- It should respect `UpdateSourceTrigger` setting.
  - `PropertyChanged` - both Text and Value changes notified immediately;
  - `LostFocus` - both Text and Value changes notified on lost focus.

### Screenshots

![NumberBoxIssue](https://github.com/lepoco/wpfui/assets/13851064/cd3c4e6a-42a7-418d-8dd9-1b15cf5b7bd5)

### OS version

```
Edition	Windows 10 Pro
Version	22H2
OS build	19045.3930
```

### .NET version

`<TargetFramework>net8.0-windows</TargetFramework>`

### WPF-UI NuGet version

`<PackageReference Include="WPF-UI" Version="3.0.0" />`

### Additional context

This is most likely related to:

- #216

but observed from the opposite end.

Issue #216, opened 2022-06-10
NumberBox not updating displayed value after changing value


dennyferra:
NumberBox does not update the displayed value (Text property) after changing Value property if the control is not empty.

**To Reproduce**  
1. Create NumberBox button 
```xml
<wpfui:NumberBox x:Name="TEST" MinWidth="100" Step="1" Min="1" Max="100" />
```
2. Create button with click event
```xml
<Button x:Name="Button" Content="Change Value" Click="Button_Click" />
```
```cs
private void Button_Click(object sender, RoutedEventArgs e) {
    TEST.Value = 20;
}
```
3. Run application and set NumberBox to 13 (or any other number)
4. Click button, notice that NumberBox still displays original number

**Expected behavior**  
I expect when changing Value for the new number to be reflected in the NumberBox

**Desktop (please complete the following information):**  
 - OS: Windows 10
 - .NET:  net6.0
 - Version: WPFUI 1.2.7

**Additional context**  
This appears to be an issue in `Value_PropertyChanged`: https://github.com/lepoco/wpfui/blob/8883c3a2a1125b75801066626e6f4d9f137dfd88/src/WPFUI/Controls/NumberBox.cs#L189

If the string is **not** null or empty the code returns. Perhaps the intention was to `return` if `IsNullOrEmpty`? I can submit a Pull Request if my assumption is correct.

In the mean time the simple workaround is to set `Text` and `Value` properties.

Thanks for the great library!

Tinekk:
The same happen with binding. The displayed value for TextBox is updating while NumberBox is not.

```
    <ui:NumberBox Value="{Binding Test}" />
    <ui:TextBox  Text="{Binding Test}" />
    <Button Command="{Binding TestIncrement}"></Button>
```

![image](https://user-images.githubusercontent.com/11279290/179953656-4d33c658-8bb9-4437-a7db-75d2f13cf858.png)

Marc-Ferrer-Castillo:
It updates when using .Text property:
`NumberBox.Text = "3";`

tmijail:
This seems to be fixed in v3.0.0

cBashTN:
I just tested v3.0.2 and doesn't seem to be fixed.

tmijail:
Here's my reproduction: https://github.com/tmijail/WPFUINumberBoxBugDemo/

At least when changing the `NumberBox`'s value in the codebehind, the issue seems fixed.


[...]

PR #1751, opened 2026-08-06
fix: numberbox text not changed

freemanke:
When we input value that not in range of minimum to maximum, the text value has not been change to the text of minimum or maximum

### DIFF
--- a/src/Wpf.Ui/Controls/NumberBox/NumberBox.cs
+++ b/src/Wpf.Ui/Controls/NumberBox/NumberBox.cs
@@ -493,6 +493,7 @@ public partial class NumberBox : Wpf.Ui.Controls.TextBox
             value = Minimum;
         }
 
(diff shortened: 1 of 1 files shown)


chucker:
Potentially related to #1315, #1016, #945, #216

Paper-Folding:
@pomianowski  What's the hold up? Please merge this already.

chucker:
I'd have to test if this doesn't have side effects.

Paper-Folding:
@chucker  Sure! 
This issue also happens when value is being clamped to min/max at previous time, at second time the value is being clamped again, however, the underlying value stays the same(clamped value), so  `FrameworkPropertyMetadata`'s `propertyChangedCallback` on `ValueProperty` is not fired, causing the text is also not updated.
A possibly better way could be using `coerceValueCallback` on `FrameworkPropertyMetadata`.

